PR #1, opened 2020-07-13
initial working version on cc

yaahc:


yaahc:
@str4d

build.rs
```diff
@@ -1,8 +1,54 @@
...
+    let out_path = env::var("OUT_DIR")?;
+    let out_path = PathBuf::from(out_path);
+    bindings
+        .write_to_file(out_path.join("bindings.rs"))
+        .wrap_err("unable to write bindings")?;
+
+    Ok(())
+}
+
+#[cfg(not(feature = "generate"))]
```
saleemrashid, line 38: I don't think it's safe to ship pre-generated bindings since bindings are generated for a specific ABI. `bindgen` is "safe" because it uses `libclang` (which we trust to be ABI-compatible with the C/C++ compiler) to query the size, alignment, etc. of types and various other ABI-dependent properties.

Hand-reviewing these bindings, they use `std::os::raw` in most places and would be largely portable for that reason. However, they include `pub type zcashconsensus_error_t = u32;` which isn't portable (since the type of an `enum` is implementation-defined).

build.rs
```diff
@@ -1,8 +1,54 @@
-extern crate cc;
+use color_eyre::{
+    eyre::{eyre, Context, Report, Result},
+    Help, SectionExt,
+};
+use std::{env, fs, path::PathBuf, process::Command};
 
-use std::env;
+trait CommandExt {
```
saleemrashid, line 7: I can't find where these extension methods are used. It looks like the last usage might have been removed in 412829b742c86ced624270a107cb74a14cc4cad0.

Cargo.toml
```diff
@@ -1,28 +1,29 @@
...
 build = "build.rs"
+edition = "2018"
 
 [lib]
-name = "bitcoinconsensus"
+name = "zcashconsensus"
 path = "src/lib.rs"
 
 [features]
-external-secp = []
```
saleemrashid, line 19: I'm not sure if this was intentionally removed (it's still referenced in the build script, and seems like a useful feature to keep around as long as it works).

src/blake2b.rs
```diff
@@ -0,0 +1,54 @@
...
+pub extern "C" fn rust_blake2b_finalize(
+    state: *mut State,
+    output: *mut c_uchar,
+    output_len: size_t,
+) {
+    let state = unsafe { state.as_mut().unwrap() };
+    let output = unsafe { slice::from_raw_parts_mut(output, output_len) };
+
+    output.copy_from_slice(state.finalize().as_bytes());
+}
```
saleemrashid, line 54: This is, as I understand it, the same file as [depend/zcash/src/rust/src/blake2b.rs](https://github.com/zcash/zcash/blob/e5300f875eca2060e233babfd70032427dea4736/src/rust/src/blake2b.rs), so it might be cleaner to share the same code.

Since the `depend/zcash` submodule is pinned to a specific commit, I don't think there's any harm in using a simple `include!` macro here. 

```suggestion
include!("../depend/zcash/src/rust/src/blake2b.rs");
```

src/lib.rs
```diff
@@ -1,212 +1,68 @@
...
+    fn it_works() {
+        let coin = i64::pow(10, 8);
+        let script_pub_key = &*SCRIPT_PUBKEY;
+        let amount = 212 * coin;
+        let tx_to = &*SCRIPT_TX;
+        let nIn = 0;
+        let flags = 1;
+        let branch_id = 0x2bb40e60;
+
+        verify_script(script_pub_key, amount, tx_to, nIn, flags, branch_id).unwrap();
```
saleemrashid, line 66: The test could also tweak one of the values and confirm that `verify_script` returns an error.

build.rs
yaahc: That sucks, I really didn't want to require CI to generate the bindings because bindgen is missing clang deps in CI :/ I'll go ahead and figure out the right command to install those

PR #22, opened 2021-06-18
Update `libzcash_script` to include support for V5 transactions

### LINKED ISSUE #21: Add support for V5 transactions
When calling `zcash_script_verify_precomputed` with a serialized V5 transaction, the library fails with a segmentation fault. This is likely because V5 transactions aren't currently fully supported.

@str4d  [mentioned](https://discord.com/channels/676527656170160146/716086297210650634/854872683736465419) that V5 transactions should be properly supported by `libzcashscript` now, so it should be a matter of updating the code to use the new version. However, [more C FFI functions might need to be added](https://discord.com/channels/676527656170160146/716086297210650634/854873388656361492).

jvff:
The upstream repository added support for [ZIP-225](https://zips.z.cash/zip-0225) and [ZIP-244](), resulting in support for V5 transactions. This PR is an attempt to close #21 by updating the upstream dependency. When I tried to update it, I ran into some issues. I'll leave this as a draft PR until these issues are solved.

The update procedure was the following:

1. Replace `depend/zcash` subtree with a submodule (see more info in the "Outstanding issues" section)
2. Include some new FFI modules required by upstream (`transaction_ffi`, `orchard_ffi` and `streams_ffi`)
3. Include `zcash/src/support/cleanse.cpp` and `zcash/src/support/cleanse.h` in the build and package
4. Replicate some upstream dependencies (`halo2`, `orchard`, `rand_core`, `tracing` and `zcash_primitives`)
5. Patch the dependency tree to use some unreleased crate versions (this must also be done on crates that use this crate, see more info in the "Outstanding issues" section)

# Outstanding issues

- [x] **subtree commit not found**: when running `git subtree pull -P depend/zcash <zcash-upstream> <commit>`, it failed with an error saying that a commit could not be found. I couldn't find the commit anywhere, but later @str4d found it and pointed out the [commit](https://github.com/str4d/zcash/commit/d94fe7678b49dce57c8a71ffe44b28d5b57eddd6) was out-of-tree. I also found the likely [rebased commit](https://github.com/zcash/zcash/commit/fa1a57867b742b0c44a9e96624e6bc58ef3912cf). I wasn't sure how to fix this using `git subtree`, so I ended up replacing the subtree with a submodule, so that the remote is properly reference to avoid similar issues in the future. It also simplified the experimentation process because I could jump to different upstream commits more easily. Long-term, this should reference an _in-tree commit in `zcash/zcash`_, independently if a subtree or a submodule is used. The submodule is currently referencing a fork because of some fixes discussed in the next item, but a reference to the fork **should NOT be** in the final PR to be merged.
  - Solution: Reference a `master` branch commit in the final PR.
- [x] **C++ compile errors**: I had to fix a few C++ compiler errors that appeared when trying to compile `libzcash_script`. These fixes can be seen [here](https://github.com/zcash/zcash/compare/master...jvff:fix-issues-with-zcash-script-crate). I'm not sure if they are correct and should be upstreamed, or if they are the consequence of a mismatched compiler version. I'm still trying to figure that out, but when discussing with @str4d, one possibility is because Zcash is compiling it as `C99` and `zcash_script` is compiling as `C++17`.
  - Solution: Compile as C99 (#23).
- [x] **patched dependencies**: currently upstream depends on some crates that don't have any releases yet, so it includes a `[patch.crates-io]` in `Cargo.toml`. That section with the patched dependencies _must_ also be included in this repository and in _any crate that uses `zcash_script`_. Ideally, these should be proper dependencies to avoid having to downstream the patches.
  - Solution: ~Wait until the patches are removed or they are more stable.~ Included as is for now. Downstream dependencies will have to apply the same patches to use this library.


teor2345:
>     * one possibility is because Zcash is compiling it as `C99` and `zcash_script` is compiling as `C++17`.

`zcash_script` was meant to be using the same language and standard version as `zcashd`. So we should fix this issue by using `C99`.

jvff:
> > * one possibility is because Zcash is compiling it as `C99` and `zcash_script` is compiling as `C++17`.
> 
> `zcash_script` was meant to be using the same language and standard version as `zcashd`. So we should fix this issue by using `C99`.

This specific issue was fixed by #23.

jvff:
I ran into a weird issue. It seems that there's a file in the Zcash source code that uses a C++20 feature: [Aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization).

https://github.com/zcash/zcash/blob/master/src/primitives/transaction.cpp#L440-L442

What's even weirder is that compiling this on Linux using C++17 works, but on Windows it doesn't. I've disabled Windows CI builds for now, but this should be fixed in the future.

.github/workflows/ci.yml
```diff
@@ -51,7 +51,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        # os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [ubuntu-latest, macOS-latest]
```
teor2345, line 55: Let's try reverting and using `language_std(&mut base_config, "c++20")` ?

```suggestion
        os: [ubuntu-latest, windows-latest, macOS-latest]
```

teor2345:
> I ran into a weird issue. It seems that there's a file in the Zcash source code that uses a C++20 feature: [Aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization).
> 
> https://github.com/zcash/zcash/blob/master/src/primitives/transaction.cpp#L440-L442
> 
> What's even weirder is that compiling this on Linux using C++17 works, but on Windows it doesn't. I've disabled Windows CI builds for now, but this should be fixed in the future.

This seems like a blocker for a Zebra release, because we'd have to disable Windows Zebra builds as well.

> For the error itself, see: https://github.com/ZcashFoundation/zcash_script/runs/3799118607#step:5:1095

```
error C7555: use of designated initializers requires at least '/std:c++20'
```

Given our tight timeframes here, I've committed the suggested fix to this PR, and re-enabled Windows builds.

build.rs
```diff
@@ -49,7 +49,7 @@ fn main() -> Result<()> {
     let target = env::var("TARGET").expect("TARGET was not set");
     let mut base_config = cc::Build::new();
 
-    language_std(&mut base_config, "c++17");
+    language_std(&mut base_config, "c++20");
```
teor2345, line 52: Try the suggested solution from the compiler:
```suggestion
    language_std(&mut base_config, "c++2a");
```

build.rs
```diff
@@ -49,7 +49,7 @@ fn main() -> Result<()> {
     let target = env::var("TARGET").expect("TARGET was not set");
     let mut base_config = cc::Build::new();
 
-    language_std(&mut base_config, "c++17");
+    language_std(&mut base_config, "c++2a");
```
teor2345, line 52: The wonderful thing about standards is that there's so many of them
```suggestion
    // try to match the standards expected by the compilers in GitHub Actions images, as of October 2021
    if target.contains("windows") {
        language_std(&mut base_config, "c++20");
    } else {
        language_std(&mut base_config, "c++17");
    }
```

build.rs
```diff
@@ -49,7 +49,12 @@ fn main() -> Result<()> {
     let target = env::var("TARGET").expect("TARGET was not set");
     let mut base_config = cc::Build::new();
 
-    language_std(&mut base_config, "c++17");
+    // try to match the standards expected by the compilers in GitHub Actions images, as of October 2021
+    if target.contains("windows") {
+        language_std(&mut base_config, "c++20");
```
teor2345, line 54:
```suggestion
        language_std(&mut base_config, "c++2a");
```

build.rs
```diff
@@ -49,7 +49,12 @@ fn main() -> Result<()> {
     let target = env::var("TARGET").expect("TARGET was not set");
     let mut base_config = cc::Build::new();
 
-    language_std(&mut base_config, "c++17");
+    // try to match the standards expected by the compilers in GitHub Actions images, as of October 2021
+    if target.contains("windows") {
+        language_std(&mut base_config, "c++2a");
```
teor2345, line 54:
```suggestion
        language_std(&mut base_config, "c++latest");
```

build.rs
```diff
@@ -49,7 +49,12 @@ fn main() -> Result<()> {
     let target = env::var("TARGET").expect("TARGET was not set");
     let mut base_config = cc::Build::new();
 
-    language_std(&mut base_config, "c++17");
+    // try to match the standards expected by the compilers in GitHub Actions images, as of October 2021
+    if target.contains("windows") {
+        language_std(&mut base_config, "c++latest");
+    } else {
+        language_std(&mut base_config, "c++17");
+    }
```
teor2345, line 57: Ok, that isn't going to work.
```suggestion
    language_std(&mut base_config, "c++17");
```

teor2345:
> > I ran into a weird issue. It seems that there's a file in the Zcash source code that uses a C++20 feature: [Aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization).
> > https://github.com/zcash/zcash/blob/master/src/primitives/transaction.cpp#L440-L442
> > What's even weirder is that compiling this on Linux using C++17 works, but on Windows it doesn't.
> 
> I assume we're using different compilers or compiler versions in the Windows and Linux tests.

We're using MSVC on Windows, and `clang` (or maybe gcc?) elsewhere.

I tried a few different standards versions, for all platforms, and specifically for Windows. But none of them worked.

I've just applied a change that always sets the compiler to `clang`.
This change doesn't add any extra dependencies, because Zebra already depends on `clang` via `bindgen`.
But it might need some `PATH` tweaks in CI, or for some users.

If this PR passes CI, let's merge, update Zebra's dependencies, and go ahead with the release.

build.rs
```diff
@@ -52,6 +52,8 @@ fn main() -> Result<()> {
     language_std(&mut base_config, "c++17");
 
     base_config
+        // use clang on all platforms, because it's already a required Zebra dependency
+        .compiler("clang")
```
teor2345, line 56: That didn't work
```suggestion
```

str4d:
> I wasn't sure how to fix this using `git subtree`,

IIRC the problem could not be fixed without rolling back the `master` branch, because the mistake was made in #16 which was merged into `master`, and repairing a broken subtree is AFAICT impossible. I did test locally, and I was able to fix the subtree in this way.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "depend/zcash"]
+	path = depend/zcash
+	url = https://github.com/jvff/zcash.git
```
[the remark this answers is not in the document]
jvff: Done in [2e4cd75](https://github.com/ZcashFoundation/zcash_script/commit/2e4cd75b0fa794af4eb0bd447ed14f0b1167d75a).

teor2345 (approved):
This is good to merge, once we're sure it works for Zebra.

daira:
> Maybe we should edit the problematic C++ file, and submit a patch to the `zcashd` repository? I'm happy to do that tomorrow. (In 14 hours time.)

Did you do this? We're not intentionally using C++20 in `zcashd`; if we did so accidentally because `clang` accepts it even in C++17 mode, that's a bug.

teor2345:
> > Maybe we should edit the problematic C++ file, and submit a patch to the `zcashd` repository? I'm happy to do that tomorrow. (In 14 hours time.)
> 
> Did you do this? We're not intentionally using C++20 in `zcashd`; if we did so accidentally because `clang` accepts it even in C++17 mode, that's a bug.
> 
> The logs at https://github.com/ZcashFoundation/zcash_script/runs/3799118607 have been deleted so I can't see which code this affected from there.

No, we split `CurrentTxVersionInfo` into another file instead, because the C++20 code was not actually used by `zcash_script`.

Here's our commit:
https://github.com/ZcashFoundation/zcash_script/pull/22/commits/2e4cd75b0fa794af4eb0bd447ed14f0b1167d75a

It still uses aggregate initialization (C++20) in the current `zcashd` code:
https://github.com/zcash/zcash/blob/f8e99e7ba501e4a3b4bdc75622788181b15129e7/src/primitives/tx_version_info.cpp#L11-21

PR #23, opened 2021-07-15
Build the `secp256k1` C files separately

jvff:
Clang currently warns that building C files using C++ mode is deprecated. This happens because the `secp256k1` C files are built together with the C++ files. This will get worse in the future because when updating the `zcash` dependency, the C files will lead to compiler errors due to subtle language version differences.

This PR solves both issues by building the `secp245k1` C files separately from the other Zcash C++ files. The resulting `libsecp256k1.a` and `libzcash_script.a` binaries are then available to be linked into the final Rust binary.

The PR also includes a few refactors to make the `main` function more concise and easier to read and update.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -60,6 +60,7 @@ fn main() -> Result<()> {
         .flag_if_supported("-Wno-reorder")
         .flag_if_supported("-Wno-deprecated-copy")
         .flag_if_supported("-Wno-unused-parameter")
+        // when compiling using Microsoft Visual C++, ignore warnings about unused arguments
         .flag_if_supported("/wd4100")
         .define("HAVE_DECL_STRNLEN", "1")
         .define("__STDC_FORMAT_MACROS", None);
--- a/build.rs
+++ b/build.rs
@@ -61,7 +61,6 @@ fn main() -> Result<()> {
         false
     };
     let target = env::var("TARGET").expect("TARGET was not set");
-    let is_big_endian = env::var("CARGO_CFG_TARGET_ENDIAN").expect("No endian is set") == "big";
     let mut base_config = cc::Build::new();
 
     base_config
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -46,77 +46,27 @@ fn bindgen_headers() -> Result<()> {
...
-        .cpp(true)
         .include("depend/zcash/src")
         .include("depend/zcash/src/rust/include/")
         .include("depend/zcash/src/secp256k1/include")
         .flag_if_supported("-Wno-implicit-fallthrough")
         .flag_if_supported("-Wno-catch-value")
         .flag_if_supported("-Wno-reorder")
         .flag_if_supported("-Wno-deprecated-copy")
+        .flag_if_supported("-Wno-unused-parameter")
+        .flag_if_supported("/wd4100")
```
teor2345, line 63: What does this warning do?
Can we add a comment here?

```suggestion
        // when compiling using Microsoft Visual C++, ignore warnings about unused arguments
        .flag_if_supported("/wd4100")
```

build.rs
```diff
@@ -142,3 +92,98 @@ fn main() -> Result<()> {
...
+        .define("ECMULT_GEN_PREC_BITS", "4")
+        // Use the endomorphism optimization now that the patents have expired.
+        .define("USE_ENDOMORPHISM", "1")
+        // Technically libconsensus doesn't require the recovery feature, but `pubkey.cpp` does.
+        .define("ENABLE_MODULE_RECOVERY", "1")
+        // The source files look for headers inside an `include` sub-directory
+        .include("depend/zcash/src/secp256k1")
+        // Some ecmult stuff is defined but not used upstream
+        .flag_if_supported("-Wno-unused-function")
+        .flag_if_supported("-Wno-unused-parameter");
```
teor2345, line 119: Do we need `/wd4100` here as well?

teor2345 (commented):
This looks good to me, but I'd like @oxarbitrage to do the final check.

build.rs
jvff: I don't think so. At least there wasn't anything from CI :man_shrugging:

oxarbitrage (approved):
It looks good to me too.

PR #34, opened 2022-05-17
Use subtree again to allow publishing the crate

conradoplg:
`cargo package` is not currently working because `cargo package` ignores subfolders with `Cargo.toml` in them. We can't work around that with submodules since they don't allow making local changes (and I couldn't get it to work in conjunction with `sparse-checkout`).

This reverts to using `subtree` (which in practice makes this PR include the entire `zcash` tree, so it looks much bigger than it is).

We were using `subtree` (exactly to solve this issue) but we had reverted to `submodule` because `subtree` behaves strangely if you use squash-and-merge. For this reason, please **do not squash-and-merge this PR** (though if that ends up happening it's easy to fix, and it's now documented in the README).

I also did not use `cargo release` for this PR because I only noticed later that we were supposed to use it. So I manually made the changes. I'm not sure if we want to keep using it from now on.


dconnolly (approved):

conradoplg:
OK, GitHub won't allow us to rebase-and-merge because `subtree` creates merge commits which are not allowed in that strategy. So I'll just squash-and-merge and we can sort this out later. I'll create a ticket for that soon

conradoplg:
https://github.com/ZcashFoundation/zcash_script/issues/35

PR #37, opened 2022-08-18
Update to zcash 5.2.0

conradoplg:
For https://github.com/ZcashFoundation/zebra/issues/4767 we need to update zcash_script to use the latest zcash sources, since we must use the same dependencies as zcash to avoid dupicates.

In 5.1.0, zcash uses cxxbridge to generate bindings in build time, which means that we must do the same. This is not ideal since we must re-do the work that `zcash` already does when compiling, but the alternative would be to build `zcash` while building this crate which is even more awkward. The long-term solution would be to move this crate to `zcash/zcash`, which we can consider later.


teor2345 (changes requested):
This looks ok, but the Windows build is failing.

Can we either fix it, or remove it and open a ticket?

conradoplg:
> This looks ok, but the Windows build is failing.
> 
> Can we either fix it, or remove it and open a ticket?

I ended up removing Windows support and opening #38

teor2345 (approved):
Thanks, looks good!

PR #48, opened 2022-12-06
Check for duplicate dependencies in CI

teor2345:
We can't use dependabot without checking dependency upgrades for duplicate dependencies.

I have already updated the branch protection rules to include these checks.


conradoplg (approved):
Looks good!

PR #60, opened 2023-02-17
Update depends/zcash to 5.4.0 release

arya2:
We want to keep zebra-script and zcash_script up to date with the latest zcash release.

Part of https://github.com/ZcashFoundation/zebra/issues/6107.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zcash_script"
-version = "0.1.7"
+version = "0.1.9"
```
arya2, line 3: Did this need to be updated or should it be left as `0.1.7`?

Cargo.toml
```diff
@@ -51,15 +51,15 @@ external-secp = []
...
+cxx = { version = "=1.0.83", features = ["c++17"] }
 libc = "0.2"
 memuse = "0.2"
 orchard = "0.3"
 rand_core = "0.6"
 tracing = "0.1"
 zcash_encoding = "0.2"
 zcash_note_encryption = "0.2"
-zcash_primitives = { version = "0.8.1", features = ["transparent-inputs"] }
+zcash_primitives = { version = "0.9.1", features = ["transparent-inputs"] }
```
conradoplg, line 62: I checked that these match https://github.com/zcash/zcash/blob/774f2d8eda62cc51990e0aa154f92f09ffba3873/Cargo.toml

However a heads up:  `zebra` is using `zcash_proofs 0.10.0` (in `zebra-consensus`) which uses `zcash_primitives 0.10.0` which will cause a duplicated dependency. Not sure how we'll want to solve that

Cargo.toml
conradoplg: We should update it, weird that it worked before :thinking:

conradoplg (approved):
Looks good, there's an issue with a dependency which we may not want to solve here, feel free to mark that comment as solved if you wish

Cargo.toml
arya2: > Not sure how we'll want to solve that

It's still using some methods that were removed in v0.10.0 so we're stuck with an item in zebra's deny.toml for now.

PR #66, opened 2023-02-23
Make dependencies automatically upgrade to match Zebra dependency versions

teor2345:
`cargo` automatically upgrades minor and patch versions to match other dependencies:
https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

But it considers `0.x` a major version, so we have to allow any upgrade to versions with a leading zero.


arya2 (approved):
Looks good.

PR #102, opened 2023-10-11
v0.1.14 release

oxarbitrage:
The versions bump are not here yet but just the upgrade from zcashd 5.6.0 to 5.7.0.

Unfortunatly there are a few small changes i had to do to the `build.rs` file for this to work. Any alternatives are welcome.


build.rs
```diff
@@ -67,13 +67,34 @@ fn gen_cxxbridge() -> Result<()> {
...
+    fs::copy(
+        "depend/zcash/src/rust/src/sapling/spec.rs",
+        "depend/zcash/src/rust/src/spec.rs",
+    )
+    .unwrap();
+    fs::copy(
+        "depend/zcash/src/rust/src/sapling/zip32.rs",
+        "depend/zcash/src/rust/src/zip32.rs",
+    )
+    .unwrap();
```
oxarbitrage, line 82: If we don't move the files inside sapling compilation will fail.

build.rs
```diff
@@ -67,13 +67,34 @@ fn gen_cxxbridge() -> Result<()> {
...
+
+    // Replace blake2s with blake2b
+    let spec = fs::read_to_string("depend/zcash/src/rust/src/spec.rs").unwrap();
+    let new_spec = spec.replace("blake2s_simd", "blake2b_simd");
```
oxarbitrage, line 86: If we don't replace the import name the compilation will fail.

oxarbitrage:
It seems the CI will not allow me to do the copying i did, i will have to find another way ...

https://github.com/ZcashFoundation/zcash_script/actions/runs/6489272390/job/17623221773?pr=102

Unsure about the clippy one as it is in the zcash code but that is a minor: https://github.com/ZcashFoundation/zcash_script/actions/runs/6489272390/job/17623221652?pr=102

build.rs
teor2345: `blake2s` and `blake2b` are different cryptographic algorithm variants, so this change could lead to consensus failures:
https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2

build.rs
teor2345: This is probably ok, but Conrado might be able to help change the import paths so this isn't needed.

teor2345:
> Unsure about the clippy one as it is in the zcash code but that is a minor: [ZcashFoundation/zcash_script/actions/runs/6489272390/job/17623221652?pr=102](https://github.com/ZcashFoundation/zcash_script/actions/runs/6489272390/job/17623221652?pr=102)

I would suggest disabling this particular clippy warning in CI using `-A clippy::(warning name)`.

build.rs
oxarbitrage: Yea, we can't do it this way as we are only supposed to change things in the `out` but nothing in the `depend/zcash`. We could of couse, disable the verification but i will like to avoid it.

build.rs
oxarbitrage: You are right there. Thank you for the feedback.

The problem here is that only `blake2b` is part of the bridge:

https://github.com/zcash/zcash/blob/v5.7.0/src/Makefile.am#L56

So i am not sure if we should have it as an external import or what to do.

oxarbitrage (commented):
Changes looks good to me, can you approve ?

conradoplg (approved):

PR #103, opened 2023-10-12
Allow clippy lint

oxarbitrage:
In the last clippy version it seems there is a new lint or an update of it `unwrap_or_default`.

Code from zcashd will trigger the lint will will make the CI fail: https://github.com/zcash/zcash/blob/v5.7.0/src/rust/src/wallet.rs#L540

This pull request will allow the lint until it is fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,4 +154,5 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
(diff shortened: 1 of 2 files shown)


conradoplg:
I think it's better to allow it in `lib.rs` so that it works everywhere and not only in our CI. Since I had already made the change to test it I went ahead and committed it, let me know if you agree with it @oxarbitrage

oxarbitrage:
Yea, sorry, i didn't knew we had that. Thanks for the fix.

PR #157, opened 2024-05-23
Remove dependencies by switching to simplified API with a callback interface

conradoplg:
zcash_script links in zcashd source code, and since the API takes a `CTransaction` that requires include a bunch of other stuff like orchard, etc.

This simplifies the API, by not passing the entire transaction but only the data required to verify. This makes this crate have ZERO Rust dependencies :tada: 

This requires using a callback to compute the sighash, since to compute it you need some information that is extracted with zcash_script.cpp.

This approach does not decrease performance because it still allows the use of precomputation in the sighash: simply precompute and store in the `ctx` pointer, which will be forwarded to the sighash callback and can be used in the sighash computation by the caller.


Cargo.toml
conradoplg, line 1: I sorted the file list and this is why the diff is weird. I know all required files are there because `cargo package` works; but feel free to double-check if there are any unneeded files in the list.

We could also delete the files in `depend/zcash` that are not here, but I think that would make it harder to update the code if needed in the future

upbqdn (approved):

PR #171, opened 2024-08-08
Provide a Rustier wrapper for zcash_script

sellout:
This adds a `Script` trait that exposes slightly Rustier types in order to have a common interface for the existing C++ implementation as well as the upcoming Rust implementation (and a third instance that runs both and checks that the Rust result matches the C++ one).


sellout:
zebra/pull/8751 integrates this with zebrad, and shouldn’t be merged until we know that that relationship is working (which it is currently not).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zcash_script"
-version = "0.2.0"
+version = "0.3.0"
```
nuttycom, line 3: We typically don't update this until we're ready to make a release, which is always done in a separate PR. If you need to refer to this from a different project, use a `[patch.crates-io]` section in that project's `Cargo.toml`

src/interpreter.rs
```diff
@@ -0,0 +1,75 @@
+bitflags::bitflags! {
+    /// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
+    ///
+    /// TODO: There are three implementations of this (with three distinct primitive types):
+    ///     - u8 constants in librustzcash,
+    ///     - i32 (well, c_int) bitflags from the C++ constants here, and
+    ///     - u32 bitflags in zebra-chain.
+    ///
+    ///   Ideally we could unify on bitflags in librustzcash.
```
nuttycom, line 9: I would suggest that we unify on the `u8` constants in `librustzcash`, with an `impl From<HashType> for c_int` instance if needed.

Cargo.toml
```diff
@@ -60,6 +60,7 @@ path = "src/lib.rs"
 external-secp = []
 
 [dependencies]
+bitflags = "2.5.0"
```
nuttycom, line 63: Ordinarily we don't pin past `major.minor` unless needed to set a lower bound.
```suggestion
bitflags = "2.5"
```

src/interpreter.rs
```diff
@@ -0,0 +1,75 @@
...
+        /// Sign one of the outputs - anyone can spend the rest
+        const Single = 3;
+        /// Anyone can add inputs to this transaction
+        const AnyoneCanPay = 0x80;
+    }
+}
+
+bitflags::bitflags! {
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    pub struct VerificationFlags: u32 {
```
nuttycom, line 25: I have verified that these match the enum values in https://github.com/ZcashFoundation/zcash_script/blob/master/depend/zcash/src/script/interpreter.h

src/zcash_script.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// A function which is called to obtain the sighash.
+///    - script_code: the scriptCode being validated. Note that this not always
+///      matches script_sig, i.e. for P2SH.
+///    - hash_type: the hash type being used.
+///
+/// The underlying C++ callback doesn’t give much opportunity for rich failure reporting, but
+/// returning `None` indicates _some_ failure to produce the desired hash.
+///
+/// TODO: Can we get the “32” from somewhere rather than hardcoding it?
```
nuttycom, line 25: Having the fixed-size array here is correct; sighashes are always 32 bytes. It would be fine to have a `#[repr(transparent)]` (or maybe `#[repr(c)]`)? newtype for the sighash though.

src/zcash_script.rs
```diff
@@ -0,0 +1,59 @@
...
+    ///  - is_final: a boolean indicating whether the input being validated is final
+    ///    (i.e. its sequence number is 0xFFFFFFFF).
+    ///  - script_pub_key: the scriptPubKey of the output being spent.
+    ///  - script_sig: the scriptSig of the input being validated.
+    ///  - flags: the script verification flags to use.
+    ///  - err: if not NULL, err will contain an error/success code for the operation.
+    ///
+    ///  Note that script verification failure is indicated by `Err(Error::Ok)`.
+    fn verify_callback(
+        sighash_callback: &SighashCallback,
```
nuttycom, line 48: I'm not sure if this can be done with a type alias, but I might write this as:
```suggestion
    fn verify_callback<SighashCallback: impl Fn(&[u8], HashType) -> Option<[u8; 32]>>(
        sighash_callback: SighashCallback,
```
This is likely to make it easier to work around lifetime errors.

Also, it's good to weaken to `FnOnce` if possible here.

nuttycom (commented):
Flushing a few initial comments.

src/zcash_script.rs
sellout: This exact change complains that “impl Fn(&[u8], HashType) -> Option<[u8; 32]>” is not a trait, but removing the `impl` at least compiles. I get a crash, though.

Also, I thought about using `FnOnce`, but I think it needs to be at least `FnMut` because the impl that compares C++ & Rust (which is not in this PR) calls the callback twice. However, I think `FnMut` is disallowed by the current C++ code, because this Rust callback is passed as the `ctx` to the underlying C++ callback, and the `ctx` is `*const c_void`.

https://github.com/ZcashFoundation/zcash_script/blob/b55a7f8f9c515c6bd9de8d05b3895e691dd2d0eb/src/lib.rs#L28-L50

There is maybe a better way to do that.

nuttycom: Oops, that `impl` shouldn't be there.

nuttycom: It's better that it be `Fn` than `FnMut` if that's possible, for sure.

src/lib.rs
```diff
@@ -1,117 +1,187 @@
...
+
+        let is_final = if is_final { 1 } else { 0 };
+
+        // SAFETY: The `script` fields are created from a valid Rust `slice`.
         let ret = unsafe {
-            super::zcash_script_verify_callback(
-                std::ptr::null(),
+            zcash_script_verify_callback(
+                (&sighash_callback as *const &SighashCallback) as *const c_void,
                 Some(sighash),
```
arya2, line 72:
```suggestion
                Some(sighash::<SighashCallback>),
```

src/lib.rs
```diff
@@ -1,117 +1,187 @@
...
-        _hash_type: c_int,
-    ) {
-        unsafe {
-            assert!(ctx.is_null());
-            let sighash =
-                hex::decode("08c7bdac77f6bb1f3aba2eaa1fada551a9c8b3b5ecd1ef86e6e58a5f1aab952c")
-                    .unwrap();
-            assert!(sighash_out_len == sighash.len() as c_uint);
+/// The sighash callback to use with zcash_script.
+extern "C" fn sighash(
```
arya2, line 29:
```suggestion
extern "C" fn sighash<SighashCallback: Fn(&[u8], HashType) -> Option<[u8; 32]>>(
```

src/zcash_script.rs
```diff
@@ -0,0 +1,59 @@
...
+/// A function which is called to obtain the sighash.
+///    - script_code: the scriptCode being validated. Note that this not always
+///      matches script_sig, i.e. for P2SH.
+///    - hash_type: the hash type being used.
+///
+/// The underlying C++ callback doesn’t give much opportunity for rich failure reporting, but
+/// returning `None` indicates _some_ failure to produce the desired hash.
+///
+/// TODO: Can we get the “32” from somewhere rather than hardcoding it?
+pub type SighashCallback = dyn Fn(&[u8], HashType) -> Option<[u8; 32]>;
```
arya2, line 26: This will also work instead of the generic closure signature.

```suggestion
pub type SighashCallback<'a> = &'a dyn Fn(&[u8], HashType) -> Option<[u8; 32]>;
```

sellout: This is perfect, thanks!

sellout:
The changes in ZcashFoundation/zebra#8751 now work correctly with this, so we’re good for final review.

src/lib.rs
```diff
@@ -1,117 +1,201 @@
...
+        script_pub_key: &[u8],
+        signature_script: &[u8],
+        flags: VerificationFlags,
+    ) -> Result<(), Error> {
+        let mut err = 0;
+
+        // SAFETY: The `script` fields are created from a valid Rust `slice`.
+        let ret = unsafe {
+            zcash_script_verify_callback(
+                (&sighash as *const SighashCalculator) as *const c_void,
```
nuttycom, line 73: I still feel like these casts are the wrong way to get this pointer.

sellout: The problem is that there’s no way to do it otherwise (AFAIK) that allows me to avoid changing the C++ code. I know it’s not sacred, but I’d like to at least get in something that works with no C++ changes first.

If I `Box`, then pointers through a bunch of C++ files lose their `const`. To me, preserving `const` seems better than switching to `Box`, but I’m certainly no expert here.

arya2: > I still feel like these casts are the wrong way to get this pointer.

It doesn't look great, but it's the same as it was in `zebra_script` (except without dereferencing the `Box`, which seems like an improvement), it's just been moved here from `zebra_script`, so if it's possible to change it, I think we should do that in another PR.

src/zcash_script.rs
arya2: Optional: Could we define a `SIGHASH_SIZE` constant here (if there isn't one elsewhere) with some documentation for why it's 32?

src/interpreter.rs
```diff
@@ -0,0 +1,70 @@
...
+        /// "At least one stack element must remain, and when interpreted as a boolean, it must be true" to
+        /// "Exactly one stack element must remain, and when interpreted as a boolean, it must be true".
+        /// (softfork safe, BIP62 rule 6)
+        /// Note: CLEANSTACK should never be used without P2SH.
+        const CleanStack = 1 << 8;
+
+        /// Verify CHECKLOCKTIMEVERIFY
+        ///
+        /// See BIP65 for details.
+        const CHECKLOCKTIMEVERIFY = 1 << 9;
```
arya2, line 68: Could we add links to any BIPs or other specifications referenced in the documentation here?

arya2 (commented):
This is looking great, thank you for making these changes.

I still need to review this in-depth, but, at a glance, I don't see any blockers.

src/lib.rs
nuttycom: I just realized what bothers me about it. This is a pointer to the callback's memory on the _stack_. That makes me uncomfortable; whenever we cross the FFI, we make sure to use heap memory.

arya2: > whenever we cross the FFI, we make sure to use heap memory

What's the advantage of using heap memory whenever we cross the FFI?

nuttycom: Giving a pointer to stack memory to a foreign process then gives that process the opportunity to choose an offset from that pointer that gives them control up the stack, which is generally a bad idea.

conradoplg: This is a function pointer though, I don't think it points to the stack. I couldn't find much information about how Rust handles Rust function pointers across FFI and back though. (For reference my original code was based on https://doc.rust-lang.org/nomicon/ffi.html#targeting-callbacks-to-rust-objects)

I'm not sure I get your point. It's the same process, not a foreign process. The C code could walk up the stack regardless.

conradoplg (commented):
It looks good, thanks! I think we just need to look a bit more into whether passing a Rust function pointer across FFI is safe.

src/cxx.rs
conradoplg, line 1: I'm a bit confused by the file structure. Is this really supposed to only have tests? Or should the Cxx implementation be moved here?

sellout: It might be a little clearer if you look at the first commit separately.
 
cxx.rs isn’t just tests – it’s the bindgen API (so, just the one `include!` line) that used to be in lib.rs, and is what’s currently used directly by Zebra (it’s still re-exported from lib.rs). The tests there are testing the bindgen functions, not the new API. Since `bindings.rs` is generated, I don’t _think_ there’s a way for me to put the tests there directly.

The other .rs files are laid out to mimic the structure of the corresponding C++ code in depend/zcash/src/script (and those get more populated in the follow-up PRs that add the actual Rust implementation).

lib.rs connects the pieces:
- reexports the bindgen API,
- contains the C++ `impl` of the new API (wrapping the bindgen API) and its tests (same tests as cxx.rs, but using the new API instead of bindgen directly), and
- _will_ contain the `impl` that runs both C++ and Rust versions of Zcash Script to compare the results (so we can ensure the entire chain behaves the same with either version).

The pure Rust `impl` will live in zcash_script.rs initially (again, to parallel depends/zcash/src/script).

conradoplg: Ah got it, thanks!

src/lib.rs
sellout: Ok, I’m sticking with the `const` for now. I’m not a fan of the raw pointers, but
1. removing `const` requires changing the C++ code, which I specifically wanted to avoid (at least initially) to ensure that there is no chance of behavioral change there),
2. `const` is Good, and
3. it’s what the original code in Zebra was doing, so again, just one less thing that differs in this PR.

@conradoplg Could you approve the workflows? Or are you not technically a maintainer, and I need someone else? (@upbqdn?)

conradoplg: Oops, forgot about it, just approve it.
I'm happy with merging the current approach but might be wise at some point to consult with someone more familiar with FFI to double check.

conradoplg (approved):

src/lib.rs
```diff
@@ -1,5 +1,201 @@
...
+    hash_type: c_int,
+) {
+    let checked_script_code_len = usize::try_from(script_code_len)
+        .expect("This was converted from a `usize` in the first place");
+    // SAFETY: `script_code` is created from a Rust slice in `verify_callback`, passed through the
+    // C++ code, eventually to `CallbackTransactionSignatureChecker::CheckSig`, which calls this
+    // function.
+    let script_code_vec =
+        unsafe { std::slice::from_raw_parts(script_code, checked_script_code_len) };
+    let ctx = ctx as *const SighashCalculator;
```
str4d, line 47: I think this should be fine for now, but we definitely will want to improve how the callbacks are handled. I don't have a good sense of how the lifetimes are being passed around.

str4d (commented):
Post-hoc review.

sellout:
> Post-hoc review.

@str4d’s review has been addressed in #175.

PR #173, opened 2024-09-25
Initial Rust implementation

sellout:
This provides two more implementations of the `ZcashScript` trait. One is a pure Rust implementation and the other runs both the C++ and Rust implementations, returning the C++ result and logging a warning if they differ.

This follows the C++ implementation extremely closely for ease of initial review. Future changes will move from this to a Rustier approach.

E.g.,
- early returns
- asserts
- bounds checking followed by accesses[^1]
- `if` instead of `match`
- mutable refs
- etc.

[^1]: accesses generally return in `Result`, so `?` ensures they’re safe, but we keep the bounds checking so that the stack state matches the C++ version (e.g., without the bounds checking, an op that pops two elements off the stack could fail after popping one, which would leave the stack in a different state than the C++ impl which checks the length of the stack before popping either).

It differs from the C++ in some places, as well. Numeric types and casting/From are changed in some places to avoid partiality. Also, the Opcodes are partitioned into multiple types because we eventually want to be able to enforce `push_only` at the type level, and `PushValue` operations carry data (although not yet in the structure), which makes them distinct from operations (`PushdataBytelength`, in particular).


sellout:
Superseded by #174, which is based on a branch on this remote, for easier chaining of PRs.

PR #174, opened 2024-09-26
Initial Rust implementation

sellout:
This provides two more implementations of the `ZcashScript` trait. One is a pure Rust implementation and the other runs both the C++ and Rust implementations, returning the C++ result and logging a warning if they differ.

This follows the C++ implementation extremely closely for ease of initial review. Future changes will move from this to a Rustier approach.

E.g.,
- early returns
- asserts
- bounds checking followed by accesses[^1]
- `if` instead of `match`
- mutable refs
- etc.

[^1]: accesses generally return in `Result`, so `?` ensures they’re safe, but we keep the bounds checking so that the stack state matches the C++ version (e.g., without the bounds checking, an op that pops two elements off the stack could fail after popping one, which would leave the stack in a different state than the C++ impl which checks the length of the stack before popping either).

It differs from the C++ in some places, as well. Numeric types and casting/From are changed in some places to avoid partiality. Also, the Opcodes are partitioned into multiple types because we eventually want to be able to enforce `push_only` at the type level, and `PushValue` operations carry data (although not yet in the structure), which makes them distinct from operations (`PushdataBytelength`, in particular).


sellout:
The files are intended to be compared side-by-side with the C++ implementation. E.g., script.rs first contains definitions from script.h, followed by definitions from script.cpp.

sellout:
@daira, @str4d, @ebfull – can’t officially add you as reviewers, but here it is.

sellout:
I don’t know why the test suite is failing on the C++ side now.
* There are no changes to the C++ in this PR.
* “It works for me” – locally running `cargo test` behaves as expected, running both the C++ and Rust tests.

My only guess is that it is somehow caused by the secp256k1(_sys) dependency maybe forcing a different version of the underlying `libsecp256k1` to be linked.

Since there’s a lot of noise in the logs, here’s the actual error:
```
  = note: /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/target/debug/build/zcash_script-13bd32f862a45584/out/libzcash_script.a(484cc929e4b13849-pubkey.o): in function `(anonymous namespace)::Secp256k1SelfTester::Secp256k1SelfTester()':
          /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:17: undefined reference to `secp256k1_selftest'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/target/debug/build/zcash_script-13bd32f862a45584/out/libzcash_script.a(484cc929e4b13849-pubkey.o): in function `CPubKey::Verify(uint256 const&, std::vector<unsigned char, std::allocator<unsigned char> > const&) const':
          /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:28: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:28: undefined reference to `secp256k1_ec_pubkey_parse'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:35: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:35: undefined reference to `secp256k1_ecdsa_signature_parse_der'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:40: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:40: undefined reference to `secp256k1_ecdsa_signature_normalize'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:41: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:41: undefined reference to `secp256k1_ecdsa_verify'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/target/debug/build/zcash_script-13bd32f862a45584/out/libzcash_script.a(484cc929e4b13849-pubkey.o): in function `CPubKey::CheckLowS(std::vector<unsigned char, std::allocator<unsigned char> > const&)':
          /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:149: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:149: undefined reference to `secp256k1_ecdsa_signature_parse_der'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:152: undefined reference to `secp256k1_context_static'
          /usr/bin/ld: /home/runner/work/zcash_script/zcash_script/depend/zcash/src/pubkey.cpp:152: undefined reference to `secp256k1_ecdsa_signature_normalize'
          collect2: error: ld returned 1 exit status
```

nuttycom (commented):
Flushing comments; reviewed up to 336d0fee4ca6f8f3956e4e9fe71d6a773198512d

src/interpreter.rs
```diff
@@ -69,3 +72,405 @@ bitflags::bitflags! {
...
+                // the stack.
+                if push_data.len() > MAX_SCRIPT_ELEMENT_SIZE {
+                    return Err(ScriptError::PushSize);
+                }
+
+                if executing {
+                    // Data is being pushed to the stack here; we may need to check
+                    // that the minimal script size was used to do so if our caller
+                    // requires it.
+                    if options.contains(VerificationFlags::MinimalData)
```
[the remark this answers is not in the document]
sellout: Do you mean because it’s not a `const` with bitwise ops?

I can push this to a later PR and swing `VerificationFlags` back to a bunch of `const`s. I think I just wrote this bit before I decided to stick as close to the C++ as possible.

src/external/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Modules that we use from Zcash, but that are outside the script directory.
```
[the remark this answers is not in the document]
sellout: These are Rust translations of files from zcashd – from the depend/zcash directory in this repo that aren’t part of zcash_script. They should go away in a later PR in favor of librustzcash or wherever else the modern equivalents live.

src/external/uint256.rs
```diff
@@ -0,0 +1,2 @@
+/// FIXME: This probably needs to be an actually separate type somewhere.
+pub type UInt256 = [u8; 32];
```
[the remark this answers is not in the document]
sellout: This is just meant to give us the minimal bit of `uint256.h` that’s needed to map things from C++ … but this type is only used in one place in the Rust code, so I’m more than happy to just use the underlying type directly, which is already used in other places, so it’d be more consistent, too.

src/script_error.rs
```diff
@@ -1,19 +1,48 @@
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(i32)]
 pub enum ScriptError {
+    Ok = 0,
```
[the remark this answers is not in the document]
sellout: The only reason for having the `repr` is for the C++/Rust comparisons, where the C++ errors get mapped to the Rust ones. If they get out of sync, the tests should fail.

src/script_error.rs
```diff
@@ -1,19 +1,48 @@
...
+    // Max sizes
     ScriptSize,
     PushSize,
     OpCount,
     StackSize,
+    SigCount,
+    PubKeyCount,
 
+    // Failed verify operations
+    VERIFY,
```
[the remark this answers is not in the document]
sellout: The ones that are all-caps map to an opcode, which are also all-caps. A better name might be something like `FailedOP_VERIFY` (boy, that doesn’t look good either) or something, but this name just matches the C++ one.

src/interpreter.rs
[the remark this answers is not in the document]
sellout: Ah, I see. This is because of the commit it’s in, which is intended to just rearrange @ebfull’s implementation a bit.

src/external/mod.rs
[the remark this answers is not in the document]
sellout: I rephrased this comment – hopefully better.

conradoplg:
> * “It works for me” – locally running `cargo test` behaves as expected, running both the C++ and Rust tests.

It works on Rust 1.81 but it doesn't on 1.82. Which is super weird, but should give you a direction where to investigate

sellout:
Yes, sorry – until I saw your comment, I didn’t realize this branch wasn’t up-to-date with my current work (I had accidentally been pushing to my fork).

I’ve updated it, fixing the merge conflicts, etc., but it still seems to have the test-building issue on Ubuntu. I’ll dig into that tomorrow.

Also, this PR still needs some review. If anyone on the ZF side (@conradoplg, @arya2, whoever) has time to give it a go, I’m happy to facilitate that however I can.

sellout:
Ok, the current failing checks are failing on master (because of Rust stable toolchain changes that happened since those commits were merged), so #176 is _trying_ to address those independently of this massive PR.

conradoplg (commented):
On a first pass looking at overall structure (but not yet the intepreter itself), this looks good, but I would love if it did not depend on `zcash_primitives` for `TxVersion` because it brings a lot of stuff with it including `orchard`, which is currently making this PR fail for some reason. We just refactored this crate to make it _not_  depend on those crates :sweat_smile: 

Some options I see:

- Create our own  `TxVersion` enum
- Don't do the "extra bits not allowed" consensus rule implementation here (which seems to be the only reason for passing the `TxVersion` round), and leave it for the caller

sellout:
> I would love if it did not depend on `zcash_primitives` for `TxVersion` because it brings a lot of stuff with it including `orchard`, which is currently making this PR fail for some reason. We just refactored this crate to make it _not_ depend on those crates 😅

Yeah, I brought up the dependency changes in https://github.com/ZcashFoundation/zcash_script/pull/175#discussion_r1781702943, when it was re-introduced.

I’ve since removed the dependency on `zcash_primitives`, which I _thought_ I had pushed to this branch, but apparently not 🙄 I’ll get that up soon.

sellout:
@conradoplg Ok, `zcash_primitives` is removed (and moved some dependencies to get the Windows test working).

src/interpreter.rs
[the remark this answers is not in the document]
nuttycom: What I mean is that https://github.com/ZcashFoundation/zcash_script/pull/174/commits/d2763f987ab81a6b39642af058516db30eac5362#diff-d1104d6a027f1f241efac6a5bc4716dfc9a757a2c3bf5258dd2746e40f86c242L106-L110 changes the logic here, so it's not a pure move.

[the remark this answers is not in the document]
nuttycom: ```git show --color-moved=dimmed-zebra --color-moved-ws=ignore-space-change d2763f987ab81a6b39642af058516db30eac5362``` is what I used to identify the deviations from pure-move.

nuttycom (commented):
Flushing a few comments.

src/interpreter.rs
```diff
@@ -262,9 +475,7 @@ pub fn eval_script(
                     stack.push_back(vch_push_value.clone());
                 }
             }
-            Operation::Constant(value) => {
-                todo!()
-            }
+            Operation::Constant(value) => stack.push_back(vec![value as u8]),
```
nuttycom: Should the payload of `Operation::Constant` just be a `u8`? The `as` cast here requires non-local reasoning.

src/interpreter.rs
```diff
@@ -545,7 +756,28 @@ pub fn eval_script(
                                 return Err(ScriptError::InvalidStackOperation);
                             }
 
-                            todo!()
+                            if let Ok(vch2) = stack.pop() {
```
nuttycom: These `if let`s are redundant with respect to the stack size check; they could just be `let ... = stack.pop()?;` instead.

src/interpreter.rs
```diff
@@ -578,25 +810,17 @@ pub fn eval_script(
...
-                                    || opcode == Opcode::OP_SHA1
-                                    || opcode == Opcode::OP_HASH160
-                                {
-                                    20
-                                } else {
-                                    32
-                                }];
-                                match opcode {
-                                    Opcode::OP_RIPEMD160 => todo!(),
+                            if let Ok(vch) = stack.pop() {
```
nuttycom: Similar thing here; can just be `let vch = stack.pop()?`

src/interpreter.rs
sellout: For this PR, I would really recommend that you do a side-by-side comparison of the full PR against the C++ code, rather than commit-by-commit. E.g., interpreter.rs should match interpreter.h followed by interpreter.cpp as closely as possible.

You’ll see that there’s not even an `Operation::Constant` in the later stages of the PR, and `as` is almost completely gone as well (with the remaining uses commented). And the changes to use `stack.pop()?` have happened as well.

Not that the history isn’t useful, but it’s mostly a path from Sean’s WIP to the full implementation, and has intermediate goals (like, “get the existing sparse C++ tests passing on the Rust impl”) that don’t really matter for the review.

nuttycom (approved):
utACK with minor suggestions. The only thing that is nominally blocking is clearing up the `// serror is set` questions

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+        };
+
+        // Hash type is one byte tacked on to the end of the signature
+        let mut vch_sig = vch_sig_in.to_vec();
+        vch_sig
+            .pop()
+            .and_then(|hash_type| HashType::from_bits(hash_type.into(), false).ok())
+            .and_then(|hash_type| (self.sighash)(script_code.0, hash_type))
+            .map(|sighash| Self::verify_signature(&vch_sig, &pubkey, &sighash))
+            .unwrap_or(false)
```
nuttycom, line 1237: This avoids the allocation of `to_vec`:
```suggestion          
       if let Some(sighash) = vch_sig_in
            .last()
            .and_then(|hash_type| HashType::from_bits((*hash_type).into(), false).ok())
            .and_then(|hash_type| (self.sighash)(script_code.0, hash_type))
       {
            Self::verify_signature(&vch_sig_in[..(vch_sig_in.len() - 1)], &pubkey, &sighash)
       } else {
            false
       }
```

src/interpreter.rs
```diff
@@ -154,24 +157,37 @@ impl CallbackTransactionSignatureChecker<'_> {
...
+        };
+
+        // Hash type is one byte tacked on to the end of the signature
+        let mut vch_sig = (*vch_sig_in).clone();
+        vch_sig
+            .pop()
+            .and_then(|hash_type| {
+                (self.sighash)(script_code.0, HashType::from_bits_retain(hash_type.into()))
+            })
+            .map(|sighash| Self::verify_signature(&vch_sig, &pubkey, &sighash))
```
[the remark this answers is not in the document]
nuttycom: This moved to https://github.com/ZcashFoundation/zcash_script/pull/174/files#r1927037595

src/interpreter.rs
```diff
@@ -605,7 +829,113 @@ pub fn eval_script(
...
+                                // distinguishable by CHECKMULTISIG NOT if the STRICTENC flag is set.
+                                // See the script_(in)valid tests for details.
+                                if !check_signature_encoding(vch_sig, flags)? {
+                                    // serror is set
+                                    return Ok(false);
+                                };
+                                check_pub_key_encoding(vch_pub_key, flags)?;
+
+                                // Check signature
+                                let ok: bool = checker.check_sig(vch_sig, vch_pub_key, script);
```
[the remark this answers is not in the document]
sellout: Yeah, the need for the consensus branch ID in the C++ code was removed with the switch to the callback API:
https://github.com/ZcashFoundation/zcash_script/blob/a9bfcd5da277a6a42dfce5dae8484ae5f50c3f34/depend/zcash/src/script/zcash_script.cpp#L56-L58

I figured it wasn’t worth duplicating dead code, but it does introduce some deviation from the C++.

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+
+                                OP_BOOLAND => ScriptNum((bn1 != bn_zero && bn2 != bn_zero).into()),
+                                OP_BOOLOR => ScriptNum((bn1 != bn_zero || bn2 != bn_zero).into()),
+                                OP_NUMEQUAL => ScriptNum((bn1 == bn2).into()),
+                                OP_NUMEQUALVERIFY => ScriptNum((bn1 == bn2).into()),
+                                OP_NUMNOTEQUAL => ScriptNum((bn1 != bn2).into()),
+                                OP_LESSTHAN => ScriptNum((bn1 < bn2).into()),
+                                OP_GREATERTHAN => ScriptNum((bn1 > bn2).into()),
+                                OP_LESSTHANOREQUAL => ScriptNum((bn1 <= bn2).into()),
+                                OP_GREATERTHANOREQUAL => ScriptNum((bn1 >= bn2).into()),
```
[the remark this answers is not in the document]
sellout: So, I don’t like the idea of that polymorphic conversion – there’s not even an `impl From<i64> for ScriptNum` yet. But there are some alternatives …

In a future commit, I get rid of the nested matching, by extracting the common logic: https://github.com/sellout/zcash_script/blob/f4dbe5fdb1ea8d7bbed966548a4e68fdc1eb9434/src/interpreter.rs#L766-L770 (not super happy with the closure syntax, though).

In the interim, one option would be to expect the result of the nested match to be a `Vec<u8>` instead of a `ScriptNum`, so the code could change like
```rust
OP_SUB =>
    (bn1 - bn2).getvch(),

OP_BOOLAND => cast_from_bool(bn1 != bn_zero && bn2 != bn_zero),
```
but … I just remembered that `cast_from_bool` isn’t in the C++ impl (and thus not in this PR). It’s something I added to avoid exactly this kind of numeric conversion later.

src/interpreter.rs
```diff
@@ -605,7 +829,113 @@ pub fn eval_script(
...
+                            let mut success = true;
+                            while success && sigs_count > 0 {
+                                let vch_sig: &ValType = stack.top(-isig as isize)?;
+                                let vch_pub_key: &ValType = stack.top(-ikey as isize)?;
+
+                                // Note how this makes the exact order of pubkey/signature evaluation
+                                // distinguishable by CHECKMULTISIG NOT if the STRICTENC flag is set.
+                                // See the script_(in)valid tests for details.
+                                if !check_signature_encoding(vch_sig, flags)? {
+                                    // serror is set
```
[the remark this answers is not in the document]
sellout: Sorry, yeah, this is confusing. It’s copied over from the C++ like this, because it’s a bug in the C++ that I’ve preserved.

The only reason this function returns `Result<bool, …>` instead of `Result<(), …>` is to preserve the C++ behavior where functions return a bool, and `false` means to use the `serror` value. The bug is that `serror` is _not_ set, and so this eventually results in `SCRIPT_ERR_UNKNOWN_ERROR`. I think this is intended to be `SCRIPT_ERR_SIG_HIGH_S` (which is currently unused). I correct that in [a future commit](https://github.com/sellout/zcash_script/blob/cc7fc610e9cb9da038cf7eb2cce083a760b60610/src/interpreter.rs#L290-L292).

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+) -> Result<bool, ScriptError> {
+    // Empty signature. Not strictly DER encoded, but allowed to provide a
+    // compact way to provide an invalid signature for use with CHECK(MULTI)SIG
+    if vch_sig.is_empty() {
+        return Ok(true);
+    };
+    if !is_valid_signature_encoding(vch_sig) {
+        return set_error(ScriptError::SigDER);
+    } else if flags.contains(VerificationFlags::LowS) && !is_low_der_signature(vch_sig)? {
+        // serror is set
```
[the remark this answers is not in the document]
sellout: This is the location of the bug I mention in https://github.com/ZcashFoundation/zcash_script/pull/174/files/243582c628d37728bb0dc8edecc83f10e02e19cc#r1927708596

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+                            // (sig pubkey -- bool)
+                            if stack.size() < 2 {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            }
+
+                            let vch_sig = stack.top(-2)?.clone();
+                            let vch_pub_key = stack.top(-1)?.clone();
+
+                            if !check_signature_encoding(&vch_sig, flags)? {
+                                //serror is set
```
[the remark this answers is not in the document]
sellout: See https://github.com/ZcashFoundation/zcash_script/pull/174/files/243582c628d37728bb0dc8edecc83f10e02e19cc#r1927708596

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+                                // to 5-byte bignums, which are good until 2**39-1, well
+                                // beyond the 2**32-1 limit of the `lock_time` field itself.
+                                let lock_time =
+                                    ScriptNum::new(stack.top(-1)?, require_minimal, Some(5))
+                                      .map_err(ScriptError::ScriptNumError)?;
+
+                                // In the rare event that the argument may be < 0 due to
+                                // some arithmetic being done first, you can always use
+                                // 0 MAX CHECKLOCKTIMEVERIFY.
+                                if lock_time < ScriptNum(0) {
```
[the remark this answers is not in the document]
sellout: Oh, there’s already a constant! I just didn’t think of it, since the C++ impl takes advantage of implicit conversion.

src/interpreter.rs
sellout: I avoid the `Vec` in [a future commit](https://github.com/sellout/zcash_script/blob/7f75bb2290ab090db212b376e61f139df5226d84/src/interpreter.rs#L267-L276) (where signature handling in general has changed). I can pull the relevant bit of that into here, though.

src/interpreter.rs
```diff
@@ -120,3 +129,1217 @@ bitflags::bitflags! {
...
+                            //     20 signatures, plus a couple other fields. u8 also gives us total
+                            //     conversions to the other types we deal with here (`isize` and `i64`).
+                            let mut i: u8 = 1;
+                            if stack.size() < i.into() {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            };
+
+                            let mut keys_count =
+                                u8::try_from(ScriptNum::new(stack.top(-isize::from(i))?, require_minimal, None)
+                                      .map_err(ScriptError::ScriptNumError)?.getint())
```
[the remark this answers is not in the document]
sellout: I added the `From` that you mentioned in https://github.com/ZcashFoundation/zcash_script/pull/174#discussion_r1925974779, which helps. The other bits change a few times in future commits to eventually look like
```rust
            let keys_count = stack
                .pop()
                .and_then(|vch| ScriptNum::new(&vch, require_minimal, None))
                .and_then(|bn| {
                    u8::try_from(bn.getint()).map_err(|err| ScriptError::PubKeyCount(Some(err)))
                })?;
```
and
```rust
            let sigs_count = stack
                .pop()
                .and_then(|vch| ScriptNum::new(&vch, require_minimal, None))
                .and_then(|bn| {
                    usize::try_from(bn.getint()).map_err(|err| ScriptError::SigCount(Some(err)))
                })?;
```
E.g., `stack.top` and `isize::from` are gone, and they’re no longer both being converted to the same type. One thought is to replace `getint` with `From` (for `i32` and larger) and `TryFrom` instances, which would help a bit more.

src/interpreter.rs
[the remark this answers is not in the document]
nuttycom: I only recommend using `From` when it's total and clear that it's safe; the important thing is that it never happens implicitly. Even though it's polymorphic, it's clear that these are boolean values that are being converted, so I don't see any issues.

src/interpreter.rs
nuttycom: Oh, yeah, `split_last` is a good choice here too.

src/script.rs
```diff
@@ -0,0 +1,589 @@
...
+
+impl Script<'_> {
+    pub fn get_op(script: &mut &[u8]) -> Result<Opcode, ScriptError> {
+        Self::get_op2(script, &mut vec![])
+    }
+
+    pub fn get_op2(script: &mut &[u8], buffer: &mut Vec<u8>) -> Result<Opcode, ScriptError> {
+        if script.is_empty() {
+            panic!("attempting to parse an opcode from an empty script");
+        }
```
conradoplg, line 455: The original seems to simply return if that happens?

```
        if (pc >= end())
            return false;
```

src/script.rs
conradoplg, line 1: I noticed that `unsigned int CScript::GetSigOpCount(const CScript& scriptSig) const` wasn't migrated - I'm guessing it's not being used here?

conradoplg (commented):
Reviewed `script.rs`, looking good so far, left a couple of comments

src/interpreter.rs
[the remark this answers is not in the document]
conradoplg: It seems very difficult to match the C++ behaviour with this approach. If `IsLowDERSignature()` returns false, then `CheckSignatureEncoding()` returns false with `serror` unchanged, but this behaviour is lost in Rust since it does not receive an input `serror`. Unless each caller of `CheckSignatureEncoding()` was checked to see this does not matter?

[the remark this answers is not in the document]
sellout: @conradoplg I agree, and it was porting to Rust that made this bug very obvious, because it is hard to parallel the behavior. But `serror` is only ever set in `set_error` (and `set_success`, which is only called once) and every use of `set_error` (except for the initialization of `serror` to `SCRIPT_ERR_UNKNOWN_ERROR`) causes an immediate return from script verification. So, keeping the `bool` and setting `Ok(false)` to `Err(ScriptError::UnknownError)` at the end seemed the closest approximation. I tried to keep the implementations close, but passing around a mutable `serror` was apparently just too much for me 😅.

In future changes (as I go into a bit in that linked comment), I get rid of the `bool`, and instead `Err(ScriptError::SigHighS)`, which I believe is the intended failure there[^1], because it makes sense and is also currently unused in the C++ impl.

[^1]: There’s a corresponding change to `lib::normalize_error` at that point, which converts `SigHighS` to `UnknownError` to make sure it compares correctly with the C++ impl.

The current C++ impl discards all `ScriptError_t` info anyway, only returning the more limited `zcash_script_error_t`, so the specifics of _which_ error is returned is lost (and `lib::normalize_error` on the Rust side discards the richer error info to make it possible to compare with the C++ results). However, to help with debugging, I have a separate branch that I haven’t pushed which has the C++ side expose `ScriptError_t` so I can check that both C++ and Rust fail for the same reason (this is particularly important for the property tests, as those will generally fail, and so the distinction between the failures is only in the errors[^2]).

[^2]: Ok, there are more unpushed branches, and those include stepwise comparison of the C++ and Rust implementations, so the state (`stack`, `op_count`, `vexec`, etc.) can be compared after each opcode to ensure both implementations are behaving identically. The reason this things are done on separate branches is because I wanted to at least initially get a Rust version up without touching the C++ version, to avoid introducing unintentional changes (e.g., the stepwise comparison uses the Rust implementation of `verify_script` and a bit deeper, then dispatches to separate C++ and Rust impls of `eval_opcode`, which doesn’t even exist in this PR).

src/script.rs
sellout: Good catch! I actually have a future commit that refactors this function to avoid mutation, and I was careful to preserve this `panic!`, because I thought it was part of the original 🤦.

I will gladly fix this!

**NB**: I think this `panic!` is unreachable in the current implementation, because all the call sites check first (in order to terminate iteration), but that’s no excuse.

src/script.rs
sellout: You’re right – it’s not used in the C++ script implementation. I’m somewhat surprised, since it looks like it’s how P2SH op counts are tracked?

nuttycom: It's only used for some mempool cost metrics that I think are no longer relevant. They might be relevant to standard rules, but I'm not sure whether they're used in Zebra.

src/interpreter.rs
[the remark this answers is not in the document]
conradoplg: The fact that we ignore the error code is a great argument for this not being a big deal. I'm happy with the current state.

conradoplg (changes requested):
Finished the review, it looks good. I'll wait for the changes you mentioned that will be pushed and overall addressing existing comments.

src/script.rs
conradoplg: Zebra doesn't currently use it. We can add it if it is ever needed.

sellout:
@conradoplg, @nuttycom: All comments should be addressed now.

src/interpreter.rs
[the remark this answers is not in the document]
sellout: I did end up adding the `From<bool> for ScriptNum` instance for now, but it disappears in a future commit when the `bool` results no longer need to pass through `ScriptNum`.

I worry about the lawlessness of things like `From` in general. I don’t think Rust’s is any worse than, say, [Haskell’s](https://hackage.haskell.org/package/witch). Sure, in this case, there is an understanding that numeric conversions are either widening (`From`) or narrowing (`TryFrom`), but then have a trait that captures those rules … and ideally the symmetries (e.g., subsets (`TryFrom<i64> for i32` + `From<i32> for i64`) vs intersections (`TryFrom<i64> for u64` + `TryFrom<u64> for i64`)). Also, the combination of them (without laws) can be misleading. Like you might expect `From<X> for Y` and `From<Y> for X` to be isomorphic when they both exist, but in this repo, that’s true for `ScriptNum` with integers in the i32 range, but we could have `From<i64> for ScriptNum` & `From<ScriptNum> for i64` and the latter (which isn’t implemented) would need to clamp the result to `i32` range. Finally, they may not be implicit, but they can be hidden (e.g., behind `?` or by any function that has a `From<X>` constraint).

src/interpreter.rs
```diff
@@ -499,9 +499,9 @@ pub fn eval_script(
                         OP_1NEGATE | OP_1 | OP_2 | OP_3 | OP_4 | OP_5 | OP_6 | OP_7 | OP_8
                         | OP_9 | OP_10 | OP_11 | OP_12 | OP_13 | OP_14 | OP_15 | OP_16 => {
                             // ( -- value)
-                            let bn =
-                                ScriptNum(i64::from(u8::from(pv)) - i64::from(u8::from(OP_1) - 1));
-                            stack.push_back(bn.getvch());
+                            let bn = ScriptNum::from(u8::from(pv))
+                                - (ScriptNum::from(u8::from(OP_1) - 1));
```
nuttycom, line 503: Is there a not already a constant that corresponds to `ScriptNum::from(u8::from(OP_1) - 1)`?

sellout: Yeah – `OP_RESERVED`. I kept it like this to mirror the C++, but I would be _more_ than happy to change that.

I assume it’s done this way because `OP_RESERVED` isn’t obviously in the position that an `OP_0` would be (if it wasn’t already somewhere else).

src/script.rs
```diff
@@ -403,6 +392,72 @@ impl ScriptNum {
...
+
+/// This is the primary extractor for numbers from `Script`, all other extractors should call this.
+impl From<ScriptNum> for i32 {
+    fn from(value: ScriptNum) -> Self {
+        if value.0 > i32::MAX.into() {
+            i32::MAX
+        } else if value.0 < i32::MIN.into() {
+            i32::MIN
+        } else {
+            value.0.try_into().unwrap()
```
nuttycom, line 436: Is the idea here that a `ScriptNum` can represent a value that is out of range internally through various computations, and is only clamped when converting back to an `i32`? In this particular case, I would have made this an `as_i32` method on `ScriptNum` instead of a `From` instance.

src/script.rs
```diff
@@ -403,6 +392,72 @@ impl ScriptNum {
...
+        i32::from(value).try_into()
+    }
+}
+
+impl From<ScriptNum> for Vec<u8> {
+    fn from(value: ScriptNum) -> Self {
+        ScriptNum::serialize(&value.0)
+    }
+}
+
```
nuttycom, line 460: This is another that I don't think should be a `From`. `ScriptNum::serialize` is adequate.

src/script.rs
sellout: Exactly. So this was previously `getint()`, but to simplify the various conversions mentioned in https://github.com/ZcashFoundation/zcash_script/pull/174#discussion_r1927796619 (and other locations), I added `TryFrom<ScriptNum> for u8`/`for u16`. I figured the `From<ScriptNum> for i32` was more consistent. However, slightly oddly, we never actually extract an `i32` from a `ScriptNum`, so reverting this instance to `getint` wouldn’t change anything other than those `TryFrom` instances.

nuttycom (approved):
utACK   fb58a44edc115676e76d6da1ccab40890272c2d1

conradoplg (approved):
Looks good!

nuttycom:
@conradoplg as a point of process, we should have waited for a third review (as consensus changes require 3 reviews). I'll recruit either @str4d or @daira to do an additional post-hoc review.

conradoplg:
> @conradoplg as a point of process, we should have waited for a third review (as consensus changes require 3 reviews). I'll recruit either @str4d or @daira to do an additional post-hoc review.

Oh sure. I'll wait for that before making a new release (we also need to test it with zebra first)

sellout:
@conradoplg I think it’d be good to get ZcashFoundation/zcash_script#190 in before a release. I just extracted it from other changes I have. It‘s small and I think it’s the only change I’ve made so far that affects the public API, so it’d avoid major version bumps down the line.

Cargo.lock
```diff
@@ -671,205 +342,178 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
- "libc",
- "rand_chacha",
- "rand_core",
+ "rand_core 0.3.1",
 ]
 
 [[package]]
-name = "rand_chacha"
-version = "0.3.1"
+name = "rand_isaac"
```
daira, line 458: Huh, this is a bit of a blast-from-the-past: [ISAAC](http://burtleburtle.net/bob/rand/isaacafa.html) is a 1996 design.

I would not use it for a CSPRNG, but according to `cargo tree -i rand_isaac` we are not doing that, so it's ok.
```
rand_isaac v0.1.1
└── rand v0.6.5
    └── proptest v0.9.6
        [dev-dependencies]
        └── zcash_script v0.2.0 ([...]/zcash_script)
```

Cargo.lock
```diff
@@ -671,205 +342,178 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
- "crossbeam-deque",
- "crossbeam-utils",
+ "autocfg 0.1.8",
+ "rand_core 0.4.2",
 ]
 
 [[package]]
-name = "reddsa"
-version = "0.5.1"
+name = "rand_xorshift"
```
daira, line 502: I checked that this is also only used via proptest:
```
$ cargo tree -i rand_xorshift
rand_xorshift v0.1.1
├── proptest v0.9.6
│   [dev-dependencies]
│   └── zcash_script v0.2.0 ([...]/zcash_script)
└── rand v0.6.5
    └── proptest v0.9.6 (*)
```

Cargo.lock
```diff
@@ -671,205 +342,178 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
+name = "rand_core"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc33ff2d4973d518d823d61aa239014831e521c75da58e3df4840d3f47749d09"
+checksum = "9c33a3c44ca05fa6f1807d8e6743f3824e8509beca625669633be0acbdf509dc"
 
 [[package]]
-name = "rand"
-version = "0.8.5"
+name = "rand_hc"
```
daira, line 449: I checked that this is only used via proptest:
```
$ cargo tree -i rand_hc
rand_hc v0.1.0
└── rand v0.6.5
    └── proptest v0.9.6
        [dev-dependencies]
        └── zcash_script v0.2.0 ([...]/zcash_script)
```

Cargo.lock
```diff
@@ -671,205 +342,178 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
- "serde",
- "thiserror",
- "zeroize",
+ "rand_core 0.3.1",
 ]
 
 [[package]]
-name = "redjubjub"
-version = "0.7.0"
+name = "rdrand"
```
daira, line 511: This is apparently only used on SGX, which is good by me because I don't trust it (it's an obvious backdoor target, even when it is working correctly; see https://en.wikipedia.org/wiki/RDRAND#Reception and https://arstechnica.com/gadgets/2019/10/how-a-months-old-amd-microcode-bug-destroyed-my-weekend/).
```
$ for target in $(rustc --print target-list); do echo ${target} && cargo tree --target ${target} -i rdrand 2>nul; done
[...]
x86_64-fortanix-unknown-sgx
rdrand v0.4.0
└── rand_os v0.1.3
    └── rand v0.6.5
        └── proptest v0.9.6
            [dev-dependencies]
            └── zcash_script v0.2.0 ([...]/zcash_script)
```

Cargo.toml
```diff
@@ -15,6 +15,7 @@ include = [
     "/README.md",
```
daira, line 15: @sellout you should add yourself to "authors".

src/external/pubkey.rs
```diff
@@ -0,0 +1,56 @@
...
+                return false;
+            };
+            // Zcash, unlike Bitcoin, has always enforced strict DER signatures.
+            if let Ok(mut sig) = ecdsa::Signature::from_der(vch_sig) {
+                // libsecp256k1's ECDSA verification requires lower-S signatures, which have
+                // not historically been enforced in Bitcoin or Zcash, so normalize them first.
+                sig.normalize_s();
+                let secp = Secp256k1::verification_only();
+                secp.verify_ecdsa(&Message::from_digest(*hash), &sig, &pubkey)
+                    .is_ok()
```
daira, line 37: This is correct for compatibility, but I can't pass up the opportunity for a rant about how much cryptographic nonsense it is.

A signature verification algorithm should be well-defined by the signature scheme itself, not on an ad hoc basis by application protocols. I get that it's easy for me to say that in hindsight, but it really should not have required hindsight. There is no advantage to either Bitcoin or Zcash to allowing non-canonical signatures in consensus and then disallowing them in standard scripts.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+}
+
+fn is_low_der_signature(vch_sig: &ValType) -> Result<bool, ScriptError> {
+    if !is_valid_signature_encoding(vch_sig) {
+        return set_error(ScriptError::SigDER);
+    };
+    // https://bitcoin.stackexchange.com/a/12556:
+    //     Also note that inside transaction signatures, an extra hashtype byte
+    //     follows the actual signature data.
+    let vch_sig_copy = vch_sig.clone();
```
daira, line 375: LIKELY CONSENSUS BUG :warning:
The C++ code is https://github.com/zcash/zcash/blob/a3435336b0c561799ac6805a27993eca3f9656df/src/script/interpreter.cpp#L176 which strips the last byte.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+    // compact way to provide an invalid signature for use with CHECK(MULTI)SIG
+    if vch_sig.is_empty() {
+        return Ok(true);
+    };
+    if !is_valid_signature_encoding(vch_sig) {
+        return set_error(ScriptError::SigDER);
+    } else if flags.contains(VerificationFlags::LowS) && !is_low_der_signature(vch_sig)? {
+        // TODO: The C++ impl claims that `serror` is set here, but it isn’t, so this eventually
+        //       ends up as `UnknownError`, when it should probably be `SigHighS`.
+        return Ok(false);
```
daira, line 406: You are correct. If `(flags & SCRIPT_VERIFY_LOW_S) != 0 && !IsValidSignatureEncoding(vchSig)`, then `IsLowDERSignature` would have set `serror` to `SCRIPT_ERR_SIG_DER`, but that case is already covered by lines 401-402 and so can't occur. Otherwise `serror` will be unset, when it should be `SCRIPT_ERR_SIG_HIGH_S`.

daira: I believe it is safe to fix this bug without causing any backward compatibility problem or consensus divergence, since Bitcoin Core did so in 2016 by https://github.com/bitcoin/bitcoin/pull/7500.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+        return opcode == OP_PUSHDATA2;
+    }
+    true
+}
+
+pub fn eval_script(
+    stack: &mut Stack<Vec<u8>>,
+    script: &Script,
+    flags: VerificationFlags,
+    checker: &dyn SignatureChecker,
```
daira, line 451: The C++ code takes a `consensusBranchId`. Is that intended to abstracted into the `SignatureChecker`? (It is only used for signature checking, I think.)

src/interpreter.rs
sellout: Ooof! Thanks for catching that!

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                                stack.push_back(vch_push_value.clone());
+                            } else {
+                                return set_error(ScriptError::BadOpcode);
+                            }
+                        }
+                    }
+                }
+            }
+            Opcode::Operation(op) => {
+                // Note how OP_RESERVED does not count towards the opcode limit.
```
daira, line 521: This comment is from the C++ code but I do not understand it. In C++, any `opcode > OP_16` *does* count toward the opcode limit, and that includes `OP_RESERVED` which is `0x50`.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                        //
+                        OP_NOP => (),
+
+                        OP_CHECKLOCKTIMEVERIFY => {
+                            // This was originally OP_NOP2 but has been repurposed
+                            // for OP_CHECKLOCKTIMEVERIFY. So, we should act based
+                            // on whether or not CLTV has been activated in a soft
+                            // fork.
+                            if !flags.contains(VerificationFlags::CHECKLOCKTIMEVERIFY) {
+                                if flags.contains(VerificationFlags::DiscourageUpgradableNOPs) {
```
daira, line 560: `SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_NOPS`, on the other hand, is only enforced as a standard rule.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                        //
+                        // Control
+                        //
+                        OP_NOP => (),
+
+                        OP_CHECKLOCKTIMEVERIFY => {
+                            // This was originally OP_NOP2 but has been repurposed
+                            // for OP_CHECKLOCKTIMEVERIFY. So, we should act based
+                            // on whether or not CLTV has been activated in a soft
+                            // fork.
```
daira, line 558:
```suggestion
                            // https://zips.z.cash/protocol/protocol.pdf#bips :
                            //
                            //   The following BIPs apply starting from the Zcash genesis block, i.e.
                            //   any activation rules or exceptions for particular blocks in the Bitcoin
                            //   block chain are to be ignored: [BIP-16], [BIP-30], [BIP-65], [BIP-66].
                            //
                            // So CHECKLOCKTIMEVERIFY is in practice always enabled (in zcashd see
                            // https://github.com/zcash/zcash/blob/a3435336b0c561799ac6805a27993eca3f9656df/src/main.cpp#L3151 ).
```

src/interpreter.rs
nuttycom: In fixing this, it would be good to add a test that fails in the "run using both" scenario.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Stack<T>(Vec<T>);
+
+/// Wraps a Vec (or whatever underlying implementation we choose in a way that matches the C++ impl
+/// and provides us some decent chaining)
+impl<T> Stack<T> {
+    fn reverse_index(&self, i: isize) -> Result<usize, ScriptError> {
+        usize::try_from(-i)
+            .map(|a| self.0.len() - a)
+            .map_err(|_| ScriptError::InvalidStackOperation)
```
daira, line 189: This allows `reverse_index(0)` which should probably be an error. It doesn't matter much, because `top(0)` will be an error.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+
+                        OP_DUP => {
+                            // (x -- x x)
+                            if stack.size() < 1 {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            }
+
+                            let a = stack.pop()?;
+                            stack.push_back(a.clone());
+                            stack.push_back(a);
```
daira, line 774: Why not more like the C++?
```suggestion
                            let vch = stack.top(-1)?;
                            stack.push_back(vch.clone());
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                        }
+
+
+                        OP_SIZE => {
+                            // (in -- in size)
+                            if stack.size() < 1 {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            }
+                            let bn =
+                                ScriptNum::try_from(stack.top(-1)?.len()).map_err(|_| ScriptError::PushSize)?;
```
daira, line 852: Is the error behaviour the same as C++ here?

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                                }
+                                let vch1 = stack.top(-2)?.clone();
+                                let vch2 = stack.top(-1)?.clone();
+                                let equal = vch1 == vch2;
+                                // OP_NOTEQUAL is disabled because it would be too easy to say
+                                // something like n != 1 and have some wiseguy pass in 1 with extra
+                                // zero bytes after it (numerically, 0x01 == 0x0001 == 0x000001)
+                                //if op == OP_NOTEQUAL {
+                                //    fEqual = !fEqual;
+                                //}
```
daira, line 876: I don't think we need to preserve comments like this.
```suggestion
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                            stack.push_back(bn.getvch())
+                        }
+
+
+                        //
+                        // Bitwise logic
+                        //
+                        OP_EQUAL
+                        | OP_EQUALVERIFY
+                        // | OP_NOTEQUAL // use OP_NUMNOTEQUAL
```
daira, line 862:
```suggestion
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                                vch_hash = Sha256::digest(vch).to_vec();
+                            } else if op == OP_HASH160 {
+                                vch_hash = Ripemd160::digest(Sha256::digest(vch)).to_vec();
+                            } else if op == OP_HASH256 {
+                                vch_hash = Sha256::digest(Sha256::digest(vch)).to_vec();
+                            }
+                            stack.pop()?;
+                            stack.push_back(vch_hash)
+                        }
+
```
daira, line 1023: I got to here in my review before my eyes started glazing over. Will resume probably on Monday.

src/interpreter.rs
conradoplg: Exactly, it was previously required because the library would compute the sighash, but now that's done by the caller via a callback.

src/interpreter.rs
nuttycom: This suggestion appears to have an error; it pushes once rather than twice.

nuttycom: That being said, I think that using `pop` is clearer; I don't know whether `.top` modifies the stack.

src/interpreter.rs
sellout: I believe it’s actually unreachable, as `ScriptError::PushSize` is what happens when a value larger than 520 bytes (`MAX_SCRIPT_ELEMENT_SIZE`) is pushed onto the stack in the first place. So `stack.top(_)?.len()` should always fit into the `i64` that `ScriptNum` holds, or `ScriptError::PushSize` would have happened already.

Should I change this to `unreachable!` with a similar explanation?

sellout (commented):
Whoops – I didn’t realize I had started a review with my comments. Sorry for not getting most of these out on Friday.

sellout:
The changes in response to post-merge comments are in #197.

src/interpreter.rs
daira: > I think that using pop is clearer [...]

Perhaps, but I asked because this PR is supposed to reflect the C++ code very closely.

src/interpreter.rs
daira: You can use `.expect` in such cases, e.g.
```suggestion
                                ScriptNum::try_from(stack.top(-1)?.len()).expect('stack element size fits in ScriptNum");
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                                return set_error(ScriptError::InvalidStackOperation);
+                            };
+
+                            let mut keys_count =
+                                u8::try_from(ScriptNum::new(stack.top(-isize::from(i))?, require_minimal, None)?)
+                                  .map_err(|_| ScriptError::PubKeyCount)?;
+                            if keys_count > 20 {
+                                return set_error(ScriptError::PubKeyCount);
+                            };
+                            op_count += keys_count;
```
daira, line 1075: On nightly we have `strict_add` which I'd suggest here. It's a little too verbose with `.checked_add` and `.expect`.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                            if stack.size() < i.into() {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            }
+
+                            let mut sigs_count =
+                                u8::try_from(ScriptNum::new(stack.top(-isize::from(i))?, require_minimal, None)?)
+                                  .map_err(|_| ScriptError::SigCount)?;
+                            if sigs_count > keys_count {
+                                return set_error(ScriptError::SigCount);
+                            };
```
daira, line 1091:
```suggestion
                          };
                          assert!(i <= 22);
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                                    success = false;
+                                };
+                            }
+
+                            // Clean up stack of actual arguments
+                            while {
+                                let res = i > 1;
+                                i -= 1;
+                                res
+                            } {
```
daira, line 1136: Since `i` is not used after this, it is equivalent to
```suggestion
                            for _ in 1..i {
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+            Some((hash_type, vch_sig)) => HashType::from_bits((*hash_type).into(), false)
+                .ok()
+                .and_then(|hash_type| (self.sighash)(script_code.0, hash_type))
+                .map(|sighash| Self::verify_signature(vch_sig, &pubkey, &sighash))
+                .unwrap_or(false),
+        }
+    }
+
+    fn check_lock_time(&self, lock_time: &ScriptNum) -> bool {
+        // There are two times of nLockTime: lock-by-blockheight
```
daira, line 1231:
```suggestion
        // There are two kinds of nLockTime: lock-by-blockheight
```
Bitcoin fixed this typo in https://github.com/bitcoin/bitcoin/commit/e846b2a1e96f27257cbb79f7f61bfaf3d873bb97

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+            return set_error(ScriptError::EvalFalse);
+        }
+    }
+
+    // The CLEANSTACK check is only performed after potential P2SH evaluation,
+    // as the non-P2SH evaluation of a P2SH script will obviously not result in
+    // a clean stack (the P2SH inputs remain).
+    if flags.contains(VerificationFlags::CleanStack) {
+        // Disallow CLEANSTACK without P2SH, as otherwise a switch CLEANSTACK->P2SH+CLEANSTACK
+        // would be possible, which is not a softfork (and P2SH should be one).
```
daira, line 1327: That was Bitcoin Core's motivation, but it doesn't make sense for Zcash because P2SH was supported from genesis.
```suggestion
        // Disallow CLEANSTACK without P2SH, because Bitcoin did.
```

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+    //        caller.
+    Ok(PubKey::check_low_s(&vch_sig_copy))
+}
+
+fn is_defined_hashtype_signature(vch_sig: &ValType) -> bool {
+    if vch_sig.is_empty() {
+        return false;
+    };
+
+    HashType::from_bits(i32::from(vch_sig[vch_sig.len() - 1]), true).is_ok()
```
daira, line 389: It is not obvious that this is equivalent to the C++, but it is.

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+        };
+
+        // Hash type is one byte tacked on to the end of the signature
+        match vch_sig_in.split_last() {
+            None => false,
+            Some((hash_type, vch_sig)) => HashType::from_bits((*hash_type).into(), false)
+                .ok()
+                .and_then(|hash_type| (self.sighash)(script_code.0, hash_type))
+                .map(|sighash| Self::verify_signature(vch_sig, &pubkey, &sighash))
+                .unwrap_or(false),
```
daira, line 1226: This doesn't seem right. In the C++, `TransactionSignatureChecker::CheckSig` calls `SignatureHash`. In the `SIGVERSION_ZIP244` case when `txTo.IsCoinBase() || txTo.vin.empty()` is false, it throws a `std::logic_error` in the case of an undefined hash_type (not 0x01, 0x02, 0x03, 0x81, 0x82, or 0x83). `HashType::from_bits((*hash_type).into(), false)`, on the other hand, will only return an error if the two least significant bits are invalid. So there is a divergence in the ZIP 244 case when an undefined bit is set in the hash type.

src/lib.rs
```diff
@@ -46,10 +51,11 @@ extern "C" fn sighash_callback(
     // function.
     let script_code_vec =
         unsafe { std::slice::from_raw_parts(script_code, checked_script_code_len) };
-    // SAFETY: `ctx` is a valid `(SighashCalculator, TxVersion)` constructed in `verify_callback`
+    // SAFETY: `ctx` is a valid `SighashCalculator` constructed in `verify_callback`
     // which forwards it to the `CallbackTransactionSignatureChecker`.
-    let (callback, tx_version) = unsafe { *(ctx as *const (SighashCalculator, TxVersion)) };
-    if let Some(sighash) = HashType::from_bits(hash_type, tx_version)
+    let callback = unsafe { *(ctx as *const SighashCalculator) };
+    // We don’t need to handle strictness here, because … something
```
daira, line 57: Please explain.

src/lib.rs
```diff
@@ -115,10 +120,131 @@ impl ZcashScript for Cxx {
...
+            lock_time,
+            is_final,
+            script_pub_key,
+            script_sig,
+            flags,
+        );
+        if rust.map_err(normalize_error) != cxx {
+            // probably want to distinguish between
+            // - C++ succeeding when Rust fails (bad),
+            // - Rust succeeding when C++ fals (worse), and
```
daira, line 212: typo

daira: Also, both are really bad because they can be used to force a consensus fork, during the time when both implementations are in use.

src/script.rs
daira: `GetLegacySigOpCount` and `GetP2SHSigOpCount` are used in zcashd's `AcceptToMemoryPool`.
`GetLegacySigOpCount` and `GetSigOpCount` are used by zcashd block construction.

The many variants of sigop counting are a mess, and I don't advocate replicating that mess in zebra. OTOH, if zebra doesn't count sigops at all then I'm concerned about potential DoS attacks.

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+    OP_HASH256 = 0xaa,
+    OP_CODESEPARATOR = 0xab,
+    OP_CHECKSIG = 0xac,
+    OP_CHECKSIGVERIFY = 0xad,
+    OP_CHECKMULTISIG = 0xae,
+    OP_CHECKMULTISIGVERIFY = 0xaf,
+
+    // expansion
+    OP_NOP1 = 0xb0,
+    OP_NOP2 = 0xb1,
```
daira, line 159: We never use `OP_NOP2`: it isn't a NOP in Zcash; it's *always* `OP_CHECKLOCKTIMEVERIFY`, so we might as well name it that and delete line 175.

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+            });
+        }
+        if require_minimal && !vch.is_empty() {
+            // Check that the number is encoded with the minimum possible
+            // number of bytes.
+            //
+            // If the most-significant-byte - excluding the sign bit - is zero
+            // then we're not minimal. Note how this test also rejects the
+            // negative-zero encoding, 0x80.
+            if (vch.last().unwrap_or_else(|| unreachable!()) & 0x7F) == 0 {
```
daira, line 295:
```suggestion
            if (vch.last().expect("not empty") & 0x7F) == 0 {

File: src/script.rs [hunk 51]
```diff
@@ -0,0 +1,642 @@
...
+                // One exception: if there's more than one byte and the most
+                // significant bit of the second-most-significant-byte is set
+                // it would conflict with the sign bit. An example of this case
+                // is +-255, which encode to 0xff00 and 0xff80 respectively.
+                // (big-endian).
+                if vch.len() <= 1 {
+                    return Err(ScriptNumError::NegativeZero);
```
daira, line 302: This returns the wrong error for `[0x00]`, since that is not a negative zero (it's a non-minimal, non-negative zero).

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+            //
+            // If the most-significant-byte - excluding the sign bit - is zero
+            // then we're not minimal. Note how this test also rejects the
+            // negative-zero encoding, 0x80.
+            if (vch.last().unwrap_or_else(|| unreachable!()) & 0x7F) == 0 {
+                // One exception: if there's more than one byte and the most
+                // significant bit of the second-most-significant-byte is set
+                // it would conflict with the sign bit. An example of this case
+                // is +-255, which encode to 0xff00 and 0xff80 respectively.
+                // (big-endian).
```
daira, line 300: It's confusing to say "big-endian" here, since the elements are not 16-bit integers.
```suggestion
                // is +-255, which encode to [0xff, 0x00] and [0xff, 0x80] respectively.
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+                if vch.len() <= 1 {
+                    return Err(ScriptNumError::NegativeZero);
+                } else if (vch[vch.len() - 2] & 0x80) == 0 {
```
daira, line 303: This returns the wrong error for `[0x00]`, since that is not a negative zero (it's a non-minimal, non-negative zero).
```suggestion
                if vch.len() == 1 && vch[0] == 0x80 {
                    return Err(ScriptNumError::NegativeZero);
                } else if vch.len() == 1 || vch[vch.len() - 2] & 0x80) == 0 {
```

src/script.rs
daira, line 300: It's confusing (arguably wrong) to say "big-endian" here, since the elements are not 16-bit integers, and the overall encoding is little-endian.
```suggestion
                // then it would have conflicted with the sign bit if one
                // fewer byte were used, and so such encodings are minimal.
                // An example of this is +-255, which have minimal encodings
                // [0xff, 0x00] and [0xff, 0x80] respectively.
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+    pub fn serialize(value: &i64) -> Vec<u8> {
+        if *value == 0 {
+            return Vec::new();
+        }
+
+        if *value == i64::MIN {
+            // The code below is buggy, and produces the "wrong" result for
+            // INT64_MIN. To avoid undefined behavior while attempting to
+            // negate a value of INT64_MIN, we intentionally return the result
+            // that the code below would produce on an x86_64 system.
```
daira, line 334:
```suggestion
            // The code below was based on buggy C++ code, that produced the
            // "wrong" result for INT64_MIN. To avoid unspecified behaviour
            // (panic or overflow) while attempting to calculate `i64::MIN.abs()`,
            // we intentionally return the result that the C++ code as compiled
            // for zcashd (with `-fwrapv`) produced on an x86_64 system.

File: src/script.rs [hunk 55]
```diff
@@ -0,0 +1,642 @@
...
+            return vec![0, 0, 0, 0, 0, 0, 0, 128, 128];
+        }
+
+        let mut result = Vec::new();
+        let neg = *value < 0;
+        let mut absvalue = value.abs();
```
daira, line 340: The C++ variable has type `uint64_t`.
```suggestion
        let mut absvalue = value.unsigned_abs();
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+
+        while absvalue != 0 {
+            result.push(
+                (absvalue & 0xff)
+                    .try_into()
+                    .unwrap_or_else(|_| unreachable!()),
```
daira, line 346:
```suggestion
                    .expect("fits in u8"),
```
(probaby needs rustfmt).

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+
+        //    - If the most significant byte is >= 0x80 and the value is positive, push a
+        //    new zero-byte to make the significant byte < 0x80 again.
+
+        //    - If the most significant byte is >= 0x80 and the value is negative, push a
+        //    new 0x80 byte that will be popped off when converting to an integral.
+
+        //    - If the most significant byte is < 0x80 and the value is negative, add
+        //    0x80 to it, since it will be subtracted and interpreted as a negative when
+        //    converting to an integral.
```
daira, line 359: Fix the weird formatting.
```suggestion
        // - If the most significant byte is >= 0x80 and the value is positive, push a
        //   new zero-byte to make the significant byte < 0x80 again.
        // - If the most significant byte is >= 0x80 and the value is negative, push a
        //   new 0x80 byte that will be popped off when converting to an integral.
        // - If the most significant byte is < 0x80 and the value is negative, add 0x80
        //   to it, since it will be subtracted and interpreted as a negative when
        //   converting to an integral.

File: src/script.rs [hunk 58]
```diff
@@ -0,0 +1,642 @@
...
+
+        if result.last().map_or(true, |last| last & 0x80 != 0) {
+            result.push(if neg { 0x80 } else { 0 });
+        } else if neg {
+            if let Some(last) = result.last_mut() {
+                *last |= 0x80;
+            }
+        }
```
daira, line 367: `value` can't be 0 because we would have returned on line 327, so the loop must have pushed at least one byte onto `result`.
```suggestion
        if result.last().expect("not empty") & 0x80 != 0 {
            result.push(if neg { 0x80 } else { 0 });
        } else if neg {
            *(result.last_mut().expect("not empty")) |= 0x80;
        }
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+
+    fn set_vch(vch: &Vec<u8>) -> Result<i64, ScriptNumError> {
+        match vch.last() {
+            None => Ok(0),
+            Some(vch_back) => {
+                if *vch == vec![0, 0, 0, 0, 0, 0, 0, 128, 128] {
+                    // On an x86_64 system, the code below would actually decode the buggy
+                    // INT64_MIN encoding correctly. However in this case, it would be
+                    // performing left shifts of a signed type by 64, which has undefined
+                    // behavior.
```
daira, line 380:
```suggestion
                    // Match the behaviour of the C++ code, which special-cased this
                    // encoding to avoided an undefined shift of a signed type by 64 bits.
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+                    return Ok(i64::MIN);
+                };
+
+                // Guard against undefined behavior. INT64_MIN is the only allowed 9-byte encoding.
```
daira, line 384:
```suggestion
                // Guard against behaviour that is not well defined. The above encoding of
                // `i64::MIN` is the only allowed 9-byte encoding.

File: src/script.rs [hunk 61]
```diff
@@ -0,0 +1,642 @@
...
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        let rhs = other.0;
+        assert!(
+            rhs == 0
+                || (rhs > 0 && self.0 <= i64::MAX - rhs)
+                || (rhs < 0 && self.0 >= i64::MIN - rhs)
+        );
+        Self(self.0 + rhs)
```
daira, line 466:
```suggestion
        Self(self.0.checked_add(other.0).expect("caller should avoid overflow"))
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+    type Output = Self;
+
+    fn sub(self, other: Self) -> Self {
+        let rhs = other.0;
+        assert!(
+            rhs == 0
+                || (rhs > 0 && self.0 >= i64::MIN + rhs)
+                || (rhs < 0 && self.0 <= i64::MAX + rhs)
+        );
+        Self(self.0 - rhs)
```
daira, line 480:
```suggestion
        Self(self.0.checked_sub(other.0).expect("caller should avoid underflow"))
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+
+    pub fn get_op2(script: &mut &[u8], buffer: &mut Vec<u8>) -> Result<Opcode, ScriptError> {
+        if script.is_empty() {
+            return Err(ScriptError::ReadError {
+                expected_bytes: 1,
+                available_bytes: 0,
+            });
+        }
+
+        // Empty the provided buffer, if any
```
daira, line 510:
```suggestion
        // Empty the provided buffer.
```

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+                    }?;
+
+                    if script.len() < size {
+                        return Err(ScriptError::ReadError {
+                            expected_bytes: size,
+                            available_bytes: script.len(),
+                        });
+                    }
+
+                    buffer.extend(&script[0..size]);
```
daira, line 550: It would be more efficient to just return `Result<(&[u8], Opcode), ScriptError>` (with appropriate lifetime on the slice).

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+                            }
+                            *script = &script[needed_bytes..];
+                            Ok(size)
+                        }
+                    };
+
+                    let size = match pv {
+                        OP_PUSHDATA1 => read_le(script, 1),
+                        OP_PUSHDATA2 => read_le(script, 2),
+                        OP_PUSHDATA4 => read_le(script, 4),
```
daira, line 539:
```suggestion
                        OP_PUSHDATA1 => usize::from(script[0]),
                        OP_PUSHDATA2 => usize::from(u16::from_le_bytes(script.try_into().unwrap())),
                        OP_PUSHDATA4 => usize::from(u32::from_le_bytes(script.try_into().unwrap())),
```
and delete `read_le`. (A further refactoring could be to put the code below in a closure so that you don't need the nested match.)

src/script.rs
```diff
@@ -0,0 +1,642 @@
...
+        }
+        assert!(opcode >= OP_1 && opcode <= OP_16);
+        (u8::from(opcode) - (u8::from(OP_1) - 1)).into()
+    }
+
+    /// Pre-version-0.6, Bitcoin always counted CHECKMULTISIGs
+    /// as 20 sigops. With pay-to-script-hash, that changed:
+    /// CHECKMULTISIGs serialized in script_sigs are
+    /// counted more accurately, assuming they are of the form
+    ///  ... OP_N CHECKMULTISIG ...
```
daira, line 592: Blech, we should get rid of this nonsense.

src/zcash_script.rs
```diff
@@ -1,16 +1,19 @@
...
 /// replicates the still-used cases, and then an `Unknown` bucket for anything else that might
 /// happen.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Error {
     /// Any failure that results in the script being invalid.
-    Ok,
+    ///
+    /// __NB__: This is in `Option` because this type is used by both the C++ and Rust
+    ///         implementations, but the C++ impl doesn’t yet expose the original error.
+    Ok(Option<ScriptError>),
```
daira, line 16: It would be better for now to just use `Ok(ScriptError::UnknownError)` for C++, I think.

daira (commented):
Reviewed with many comments!

src/script.rs
sellout: With the other changes that I backported, this is now handled with a nice `split_at_checked`.

src/script.rs
sellout: This is meant as a suggestion for the future, right? Because this would lead to a consensus fork, since the C++ version would reject blocks that Rust sees as valid.

sellout (commented):
Finally addressed @daira’s review. The changes are all up at #197.

Cargo.toml
daira: A :tada: means that the comment has been addressed in https://github.com/ZcashFoundation/zcash_script/pull/197

src/interpreter.rs
daira: (This is fixed by #197 and is not a potential consensus divergence between zcashd and zebra, so we can discuss it in public.)

I think the above reasoning is incorrect. Let's say that `foo` is a valid signature but `foo` with the last byte stripped is not. If we pass `foo` to libsecp256k1, it will pass, but stripping the last byte would have made it fail.

The point here is that consensus divergence in both directions is important. We cannot validly reason "this divergence doesn't matter because if it was applied to a valid state, it would still be valid". We also have to consider what happens if the divergence is applied to an invalid state: can it incorrectly make that state valid? In this case the divergence is to add an extra byte to the signature, which *can* turn it from invalid to valid.

src/interpreter.rs
daira: A :confused: emoji indicates that this is *not* addressed in #197. (It's not a great choice of emoji, there just isn't a better one in the supported set.)

src/interpreter.rs
```diff
@@ -120,3 +129,1207 @@ bitflags::bitflags! {
...
+                            if stack.size() < i.into() {
+                                return set_error(ScriptError::InvalidStackOperation);
+                            };
+
+                            let mut keys_count =
+                                u8::try_from(ScriptNum::new(stack.top(-isize::from(i))?, require_minimal, None)?)
+                                  .map_err(|_| ScriptError::PubKeyCount)?;
+                            if keys_count > 20 {
+                                return set_error(ScriptError::PubKeyCount);
+                            };
```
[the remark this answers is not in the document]
daira: I had checked that the assertion is valid.

src/interpreter.rs
daira: 🤔

src/lib.rs
```diff
@@ -111,6 +114,93 @@ impl ZcashScript for Cxx {
     }
 }
 
+/// Runs both the C++ and Rust implementations `ZcashScript::legacy_sigop_count_script` and returns
+/// both results. This is more useful for testing than the impl that logs a warning if the results
+/// differ and always returns the C++ result.
```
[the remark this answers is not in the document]
daira: 🤔

src/lib.rs
daira: 🤔

src/script.rs
daira: Yes, that was my intention. It would have to be done carefully, but I believe it's not consensus.

src/interpreter.rs
sellout: Right, but that direction is also checked – `is_valid_signature_encoding` checks that the signature is a valid DER signature, _plus_ an extra byte for the hash type, so we can’t end up with the case where the hash type is the same byte as the last byte of the signature.

But … all of this is non-local, so it’s hard to keep track of at best. The future commits hopefully make that better.

src/interpreter.rs
sellout: So, I don’t think there’s anything to do here – this matches the callback-based C++ as used by Zebra.

conradoplg: I believe that needs to be enforced by the caller, in the callback, since this crate is not aware of whether it's a NU5-onward tx. I'll make a note to double check this in Zebra.

sellout: It’s on the caller to include the `StrictEnc` flag for NU5 tx[^1], it doesn’t need to be handled in the callback (and can’t be because the callback receives a `HashType`, not the bits)[^2].

If that flag is set, then `from_bits` gets called with `is_strict: true` from https://github.com/ZcashFoundation/zcash_script/blob/2afc4743383686411941fba431f1bf3a8271c059/src/interpreter.rs#L386-L392 which causes https://github.com/ZcashFoundation/zcash_script/blob/2afc4743383686411941fba431f1bf3a8271c059/src/interpreter.rs#L61-L63 to fail if any bits other than the three we care about are set.

It’s annoying to follow in this match-the-C++ PR, because each call to `from_bits` has `is_strict` hardcoded, but [in a future commit](https://github.com/sellout/zcash_script/blob/5efdcecf920433a207110eb6b46f8287e28505fd/src/interpreter.rs#L274-L291) `is_strict` is set directly from the `StrictEnc` flag.

[^1]: I forget if it even made it into master, but there was a brief period when I reintroduced `TxVersion` (and a dependency on zcash_primitives) here, so that NU5 transactions would be strict regardless of whether `StrictEnc` was set, but that was ultimately shot down.

[^2]: The wrapper used to pass a Rust callback to the C++ implementation explains why it’s ok to be non-strict there https://github.com/ZcashFoundation/zcash_script/blob/2afc4743383686411941fba431f1bf3a8271c059/src/lib.rs#L131-L134

PR #175, opened 2024-09-27
Address Str4d’s comments on #171

sellout:
Notably, `HashType` has changed incompatibly, so
ZcashFoundation/zebra#8751 will need to be updated.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1,18 +1,34 @@
-bitflags::bitflags! {
-    /// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
-    ///
-    /// TODO: This is currently defined as `i32` to match the `c_int` constants in this package, but
-    ///       should use librustzcash’s `u8` constants once we’ve removed the C++.
-    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-    pub struct HashType: i32 {
-        /// Sign all the outputs
-        const All = 1;
-        /// Sign none of the outputs - anyone can spend
-        const None = 2;
-        /// Sign one of the outputs - anyone can spend the rest
-        const Single = 3;
-        /// Anyone can add inputs to this transaction
-        const AnyoneCanPay = 0x80;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SignedOutputs {
+    /// Sign all the outputs
+    All,
+    /// Sign one of the outputs - anyone can spend the rest
+    Single,
+    /// Sign none of the outputs - anyone can spend
+    None,
+}
+
+/// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct HashType {
+    pub signed_outputs: SignedOutputs,
+    /// Anyone can add inputs to this transaction
+    pub anyone_can_pay: bool,
+}
+
+impl HashType {
+    pub fn from_bits(bits: i32) -> Option<Self> {
+        let msigned_outputs = match (bits & 2 != 0, bits & 1 != 0) {
+            (false, false) => None,
+            (false, true) => Some(SignedOutputs::All),
+            (true, false) => Some(SignedOutputs::None),
+            (true, true) => Some(SignedOutputs::Single),
+        };
+
+        msigned_outputs.map(|signed_outputs| HashType {
+            signed_outputs,
+            anyone_can_pay: bits & 0x80 != 0,
+        })
     }
 }
 
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1,3 +1,5 @@
+use zcash_primitives::transaction::TxVersion;
+
 /// The ways in which a transparent input may commit to the transparent outputs of its
 /// transaction.
 ///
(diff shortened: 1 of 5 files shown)


src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
+pub enum SignedOutputs {
+    /// Sign all the outputs
+    All,
+    /// Sign one of the outputs - anyone can spend the rest
+    Single,
+    /// Sign none of the outputs - anyone can spend
+    None,
+}
+
+/// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
```
str4d, line 11: Other useful references:
- https://zips.z.cash/zip-0244#s-2a-hash-type
- https://zips.z.cash/protocol/protocol.pdf#sighash

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
+            (false, true) => Some(SignedOutputs::All),
+            (true, false) => Some(SignedOutputs::None),
+            (true, true) => Some(SignedOutputs::Single),
+        };
+
+        msigned_outputs.map(|signed_outputs| HashType {
+            signed_outputs,
+            anyone_can_pay: bits & 0x80 != 0,
+        })
     }
```
str4d, line 32: This constructor ignores the value of the undefined bits. However, per the reference to the protocol spec that I gave above, for v5 transactions there is a consensus rule that the sighash type is a canonical encoding, so those other bits must be zero:
> [NU5 onward] Any SIGHASH type encoding used in a version 5 transaction **MUST** be the canonical encoding of one of the defined SIGHASH types, i.e. one of `0x01`, `0x02`, `0x03`, `0x81`, `0x82`, or `0x83`. (Previously, undefined bits of a SIGHASH type encoding were ignored.)

This constructor should be given an argument `require_canonical: bool` that lets the caller specify the behaviour it requires.

Also, please document this method.

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
+    Single,
+    /// Sign none of the outputs - anyone can spend
+    None,
+}
+
+/// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct HashType {
+    pub signed_outputs: SignedOutputs,
+    /// Anyone can add inputs to this transaction
```
str4d, line 15:
```suggestion
    /// Allows anyone to add transparent inputs to this transaction.
```

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
-        const All = 1;
-        /// Sign none of the outputs - anyone can spend
-        const None = 2;
-        /// Sign one of the outputs - anyone can spend the rest
-        const Single = 3;
-        /// Anyone can add inputs to this transaction
-        const AnyoneCanPay = 0x80;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SignedOutputs {
+    /// Sign all the outputs
```
str4d, line 3:
```suggestion
    /// The input signature commits to all transparent outputs in the transaction.
```

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
+    All,
+    /// Sign one of the outputs - anyone can spend the rest
```
str4d, line 5:
```suggestion
    /// The transparent input's signature commits to the transparent output at the same
    /// index as the transparent input.
    ///
    /// If the specified transparent output along with any shielded outputs only consume
    /// part of this input, anyone is permitted to modify the transaction to claim the
    /// remainder.
```

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
+    Single,
+    /// Sign none of the outputs - anyone can spend
```
str4d, line 7:
```suggestion
    /// The transparent input's signature does not commit to any transparent outputs.
    ///
    /// If the shielded outputs only consume part (or none) of this input, anyone is
    /// permitted to modify the transaction to claim the remainder.
```

src/interpreter.rs
```diff
@@ -1,18 +1,34 @@
...
-        /// Sign all the outputs
-        const All = 1;
-        /// Sign none of the outputs - anyone can spend
-        const None = 2;
-        /// Sign one of the outputs - anyone can spend the rest
-        const Single = 3;
-        /// Anyone can add inputs to this transaction
-        const AnyoneCanPay = 0x80;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum SignedOutputs {
```
str4d, line 2:
```suggestion
/// The ways in which a transparent input may commit to the transparent outputs of its
/// transaction.
///
/// Note that:
/// - Transparent inputs always commit to all shielded outputs.
/// - Shielded inputs always commit to all outputs.
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
pub enum SignedOutputs {
```

str4d (changes requested):
Reviewed e748f2fca1a7a678d01f5b20480d28795146bf3d.

src/interpreter.rs
sellout: I‘ve made these changes – although rather than `require_canonical: bool`, I added `tx_version: zcash_primitives::transaction::TxVersion`.

However, in the current (C++) implementation, the `extern "C"` callback is passed the bitflags, and it converts it to `HashType` before passing it to the Rust callback provided by Zebra. The problem here is that we don’t know the tx version in the `extern "C"` callback.

One option is to pass the bitflags along to the Rust callback, allowing Zebra (or whatever consumer) to check it against the tx version, which eliminates the need for `HashType` in zcash_script[^1], but [the current Zebra doesn’t check this](https://github.com/ZcashFoundation/zebra/pull/8751/files#diff-eecd7497f79586adb8aeebb4559fc9fbc14d085c6d50e25750f9da17d24ff462L121), which I think is a good argument that we shouldn’t push the responsibility out to the consumer.

[^1]: #174 _does_ use `HashType`, but in its C++ parallel structure, it only extracts bits from it, so if there is no other use of the structure, that can just use the bits directly. Later iterations of the Rust implementation will care about `HashType` more.

Another option is to pass the tx version into script validation somewhere, and then the `extern "C"` callback `ctx` can be a tuple of the Rust callback and the tx version, so we can build the `HashType` correctly. I like this version the best, but I _think_ the reason for introducing the callback in the first place was to remove the dependence on tx information from script validation, so maybe this is rolling things back?

A third option is to always pass `TxVersion::Sapling` to `HashType::from_bits` in the callback (opening an associated issue & adding a comment referencing it), which would match the existing (lax) Zebra behavior.

I’m implementing the middle one, but let me know if you think there’s a better option.

str4d (approved):
utACK b7bdbc91cbd937d17a19bbea02a70a981a3b7487. I'm a little iffy about the current callback API, but I think it's fine for now and we can iterate as things progress.

conradoplg (approved):

src/interpreter.rs
conradoplg: This seems fine for now, the motivation for the callback was to prevent C++ code from having to parse transactions. This doesn't change that so it's OK.

conradoplg (approved):

PR #176, opened 2024-12-06
Add a basic rust-toolchain.toml

sellout:
Matches the one in Zebra, although I think pinning to a specific release (e.g., `channel = "1.82"`) would help with consistency, perhaps avoiding issues like
https://github.com/ZcashFoundation/zcash_script/pull/174#issuecomment-2445336306

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,18 +10,10 @@ jobs:
   check:
     name: Check
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        rust:
-          - stable
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: true
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          override: true
       - uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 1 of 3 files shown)


sellout:
Ah, good. This captures that the build failures are due to rustc 1.81, and not anything in #174. And also a reason to pin to a specific rustc version – changes outside of the repo shouldn’t cause a working repo to start failing.

sellout:
@conradoplg The issue from https://github.com/ZcashFoundation/zcash_script/pull/174#issuecomment-2427737764 occurs on master, this PR tries to pin down rustc a bit to avoid it.

- 1.81 – works
- 1.82–1.84 (beta) – not working
- 1.85 (nightly) – working again

So, the problem seems to be fixed with just a bit of waiting … however I can’t find anything obvious in the 1.82 or 1.85 changelogs that would indicate why this started/stopped happening.

I’ve currently pinned this to 1.81 (and I would ideally pin to a specific version even if there wasn’t a bug, because it’s better for consistency). Zebra would need to be similarly pinned for it to depend on this before mid-Feb.

However (as you can see in the failing check) `unsafe extern "C"` wasn’t supported until 1.82, and `cargo package` is complaining about it. But I don’t understand why [master passed the same check when it used 1.81](https://github.com/ZcashFoundation/zcash_script/actions/runs/11114320609/job/30880543011?pr=175) (and there are no changes related to that code in this PR).

I think everything would pass if I pinned to 1.85, but
1. I can’t actually pin to `"1.85"` since it’s not released yet, and
2. I don’t want to pin to `"nightly"` because it’ll keep shifting.

Maybe I can pin to today’s date?

sellout:
> However (as you can see in the failing check) `unsafe extern "C"` wasn’t supported until 1.82, and `cargo package` is complaining about it. But I don’t understand why [master passed the same check when it used 1.81](https://github.com/ZcashFoundation/zcash_script/actions/runs/11114320609/job/30880543011?pr=175) (and there are no changes related to that code in this PR).

I understand why this failure is happening. The `unsafe extern` is generated by bindgen, which targets latest stable by default. When 1.81 was current, it would have generated `extern` instead of `unsafe extern`. Simply adding a rust-toolchain.toml doesn’t affect bindgen, so it was generating 1.83-compatible code, which didn’t work with our rust-toolchain.toml pinning to 1.81.

What’s confusing me now is that the Cargo.lock file pins a version of bindgen (0.69.4) that is too old to generate rustc 1.82-specific code, but _one_ of the CI jobs (“Package (rust_toolchain.toml)”) appears to ignore the Cargo.toml and pull a newer version of bindgen that _does_ support 1.82, which is why that was the only job that was failing.

For now, I am setting the `rust_target` to the newest version supported by bindgen 0.69.4, which is 1.73, and quieting the deprecation error that the one CI job triggers. But I would love to have that CI job not get a different bindgen version than everything else …

nuttycom (approved):
utACK f6857f1692c8df11d7ccf224d9f0411afa96e41e

conradoplg:
> What’s confusing me now is that the Cargo.lock file pins a version of bindgen (0.69.4) that is too old to generate rustc 1.82-specific code, but _one_ of the CI jobs (“Package (rust_toolchain.toml)”) appears to ignore the Cargo.toml and pull a newer version of bindgen that _does_ support 1.82, which is why that was the only job that was failing.
> 
> For now, I am setting the `rust_target` to the newest version supported by bindgen 0.69.4, which is 1.73, and quieting the deprecation error that the one CI job triggers. But I would love to have that CI job not get a different bindgen version than everything else …

I'm not sure, but my guess is that `cargo package` ignores the lock file since it's a library. So when it tries to compile the crate, it will use the most recent version that obeys the constraints in `Cargo.toml`.

I don't think we can afford to pin absolutely everything because this is a library and we can't really dictate the specific versions of all dependencies and the Rust version that library users will have.

This seems good for now (sorry for taking so long to chime in), let's hope everything works on 1.85 and then we can remove the workarounds.

PR #187, opened 2025-01-15
Bump bindgen 0.71.1

conradoplg:
https://github.com/ZcashFoundation/zcash_script/pull/179 was failing because proc-macro2 needs to be update in the lock file (probably some crate in the import path it's not specifying a correct minimum version)


nuttycom (approved):
utACK d95515f90af93b4019624acaadf8c924b16d58f3

PR #188, opened 2025-01-22
Fix clippy warnings

rex4539:


src/lib.rs
```diff
@@ -130,15 +130,15 @@ mod tests {
             .unwrap()
             .as_slice()
             .first_chunk::<32>()
-            .map(|hash| *hash)
+            .copied()
```
sellout, line 133: Perhaps outside the scope of silencing Clippy warnings, but both of these expressions could be simplified to
```rust
        <[u8; 32]>::from_hex("e8c7bdac77f6bb1f3aba2eaa1fada551a9c8b3b5ecd1ef86e6e58a5f1aab952c")
            .ok()

```

rex4539: Feel free to apply the fix you are most comfortable with :)

nuttycom (approved):
utACK cd62fd12fd68646aab9cafa31ab89804ebdc5f96

PR #190, opened 2025-01-30
Change type of `lock_time` parameter

sellout:
This is a breaking change. Lock times are stored in tx as `u32`, but this API expected `i64`, forcing conversions on the caller. This change brings the API into alignment with the tx representation.


nuttycom (dismissed):
utACK 9cf1dcf0631ffa1e22cf1c3e2aaec13a6ce498f1

conradoplg (approved):
Looks good

PR #195, opened 2025-01-30
Expose `ScriptError` on C++ side

sellout:
**This changes the C++ implementation.**

Provides richer errors, which also gives more precision when comparing against the Rust impl.

This also removes the (now unused) `zcash_script_error_t`. The only case other than `zcash_script_ERR_OK` that was still in use was `zcash_script_ERR_VERIFY_SCRIPT`, so that case has been added to `ScriptError`.

This avoids changing the Rust API, but potentially `Error` and `ScriptError` on the Rust side could be collapsed into one `enum`. It would just be a breaking change.

### DIFF
--- a/depend/zcash/src/script/script_error.h
+++ b/depend/zcash/src/script/script_error.h
@@ -53,6 +53,8 @@ typedef enum ScriptError_t
     /* softfork safeness */
     SCRIPT_ERR_DISCOURAGE_UPGRADABLE_NOPS,
 
+    SCRIPT_ERR_VERIFY_SCRIPT,
+
     SCRIPT_ERR_ERROR_COUNT
 } ScriptError;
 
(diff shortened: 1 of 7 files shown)


sellout:
Since property/fuzz tests generally fail, this change was crucial to thorough testing of #174. Without it, there are only two failure values from the C++ side.

This is also preliminary work for the deeper step-wise comparison between the interpreters, as `ScriptError` is the failure type within verification.

conradoplg (approved):
Looks great, thanks. `zcash_script_error` did became useless after the previous refactoring.

> This avoids changing the Rust API, but potentially Error and ScriptError on the Rust side could be collapsed into one enum. It would just be a breaking change.

I think this is worth doing, but feel free to do it in a separate PR.

I'll leave this open for a bit for other reviewers to chime in.

PR #197, opened 2025-02-03
Addressing post-hoc PR feedback on #174

sellout:


nuttycom (dismissed):
utACK 9ab842671910ee51be3d242fe2a2e274e52b9f21

conradoplg (commented):
This is looking good but I think that @daira's latest comments on https://github.com/ZcashFoundation/zcash_script/pull/174 haven't been addressed yet. Also needs solving conflicts.

sellout:
> This is looking good but I think that @daira's latest comments on #174 haven't been addressed yet. Also needs solving conflicts.

Just finished both of those – and moved this out of draft.

conradoplg (dismissed):
LGTM; I'll wait for @daira's review before merging

src/script.rs
```diff
@@ -595,10 +547,11 @@ impl Script<'_> {
         let mut pc = self.0;
         let mut last_opcode = Opcode::Operation(OP_INVALIDOPCODE);
         while !pc.is_empty() {
-            let opcode = match Self::get_op(&mut pc) {
+            let (opcode, new_pc) = match Self::get_op(pc) {
                 Ok(o) => o,
                 Err(_) => break,
```
nuttycom, line 552: nonblocking: It's annoying that the error case here is ignored in the original C++ code. It will probably be worth a comment here, since once we transition fully to the Rust implementation this behavior will be puzzling.

nuttycom (approved):
utACK 9ab842671910ee51be3d242fe2a2e274e52b9f21

src/script.rs
sellout: Yeah … I can only guess that this is done to prevent someone from creating script that’s too expensive, but then doing something invalid at the end to prevent this check from catching it. _But_
1. the interpreter also tracks op_count, so it at least wouldn’t get past 201 no matter what; and
2. if this _did_ error, then we’d know not to bother running the script. 🤷🏼

So, should the comment just be something like “TODO: `Err` is ignored to match the C++ behavior. In future, allow this operation to fail.”?

src/interpreter.rs
```diff
@@ -372,13 +372,15 @@ fn is_low_der_signature(vch_sig: &ValType) -> Result<bool, ScriptError> {
     // https://bitcoin.stackexchange.com/a/12556:
     //     Also note that inside transaction signatures, an extra hashtype byte
     //     follows the actual signature data.
-    let vch_sig_copy = vch_sig.clone();
+    let (_, vch_sig_copy) = vch_sig.split_last().unwrap_or_else(|| {
+        unreachable!("`is_valid_signature_encoding` checks that the length is at least 9.")
+    });
```
daira, line 377:
```suggestion
    let (_, vch_sig_copy) = vch_sig
        .split_last()
        .expect("`is_valid_signature_encoding` has checked that the length is at least 9");
```
In general `.unwrap_or_else(|| unreachable!(msg))` should instead be `.expect(msg)`. The latter means the same thing, but is shorter. (Don't worry about rustfmt sometimes making code _look_ more verbose.)

src/script.rs
daira: "... In future, consider allowing this operation to fail." (I haven't checked whether that is consensus-compatible.)

daira (approved):
utACK. Search for 🤔 in https://github.com/ZcashFoundation/zcash_script/pull/174/files to see comments there that are *not* addressed here.

daira (approved):
re-utACK. My remaining comments are non-blocking.

sellout:
Since one of the requirements on this repo is that all comments must be resolved, I pushed two more commits that address the outstanding comments here and on #174, respectively.

nuttycom (approved):
re-utACK 2d835f6bc6337404b7e68efb1ed551d636d2f7bb

conradoplg (approved):
Looks good!

PR #204, opened 2025-03-11
Add various stepwise functionality

sellout:
This extracts an `eval_step` from the Rust interpreter’s `eval_script` and adds tooling around that.

Most significantly, this adds a `StepFn` trait with implementations for
- standard Rust evaluation,
- “broken” evaluation (in the current example, it drops `OP_EQUAL` from scripts, to trigger divergences), and
- stepwise comparison of evaluators.

The latter will be used to do deep comparisons with the C++ interpreter when the `StepFn` impl for C++ is added in a future PR.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -483,11 +483,11 @@ impl State {
 
     /// Creates a state with an initial stack, but other components empty.
     pub fn initial(stack: Stack<Vec<u8>>) -> Self {
-        Self::manual(stack, Stack::new(), 0, Stack::new())
+        Self::from_parts(stack, Stack::new(), 0, Stack::new())
     }
 
-    /// Create an arbitrary stat.
-    pub fn manual(
+    /// Create an arbitrary state.
+    pub fn from_parts(
         stack: Stack<Vec<u8>>,
         altstack: Stack<Vec<u8>>,
         op_count: u8,
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1,4 +1,3 @@
-use std::mem::swap;
 use std::slice::Iter;
 
 use ripemd::Ripemd160;
@@ -144,6 +143,7 @@ pub struct BaseSignatureChecker();
 
 impl SignatureChecker for BaseSignatureChecker {}
 
+#[derive(Copy, Clone)]
 pub struct CallbackTransactionSignatureChecker<'a> {
     pub sighash: SighashCalculator<'a>,
     pub lock_time: &'a ScriptNum,
@@ -182,7 +182,7 @@ pub struct Stack<T>(Vec<T>);
 
 /// Wraps a Vec (or whatever underlying implementation we choose in a way that matches the C++ impl
 /// and provides us some decent chaining)
-impl<T> Stack<T> {
+impl<T: Clone> Stack<T> {
     fn reverse_index(&self, i: isize) -> Result<usize, ScriptError> {
         usize::try_from(-i)
             .map(|a| self.0.len() - a)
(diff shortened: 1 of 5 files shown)


sellout:
**NB**: I strongly recommend review at least the interpreter changes with whitespace hidden – the changes there are minimal, but because it was extracted to a new function, indentation changed.

src/lib.rs
```diff
@@ -424,15 +470,26 @@ mod tests {
             is_final in prop::bool::ANY,
             pub_key in prop::collection::vec(0..=0xffu8, 0..=OVERFLOW_SCRIPT_SIZE),
             sig in prop::collection::vec(0..=0xffu8, 1..=OVERFLOW_SCRIPT_SIZE),
-            flags in prop::bits::u32::masked(VerificationFlags::all().bits()),
+            flag_bits in prop::bits::u32::masked(VerificationFlags::all().bits()),
         ) {
-            let ret = check_verify_callback::<CxxInterpreter, RustInterpreter>(
-                &missing_sighash,
+            let flags =                 repair_flags(VerificationFlags::from_bits_truncate(flag_bits));
```
daira, line 479: Odd formatting.

src/lib.rs
```diff
@@ -18,11 +18,11 @@ use std::os::raw::{c_int, c_uint, c_void};
...
 
-pub use cxx::*;
 pub use interpreter::{
     CallbackTransactionSignatureChecker, DefaultStepEvaluator, HashType, SighashCalculator,
     SignedOutputs, VerificationFlags,
 };
+use script_error::ScriptError;
 pub use zcash_script::*;
 
 /// A tag to indicate that the C++ implementation of zcash_script should be used.
```
daira, line 28: This is not a tag.

daira (approved):
utACK with nonblocking comments.

nuttycom (commented):
Flushing a few comments from the first couple of commits.

src/zcash_script.rs
```diff
@@ -163,83 +158,53 @@ impl<'a, T, U> StepFn for ComparisonStepEvaluator<'a, T, U> {
     }
 }
 
-impl ZcashScript for RustInterpreter {
+pub struct StepwiseInterpreter<F>
+where
+    F: StepFn,
+{
+    pub initial_payload: F::Payload,
+    pub stepper: F,
```
nuttycom, line 166: Do these need to be public?

nuttycom (changes requested):
Questions about `pub` visibility of bare struct members are blocking otherwise LGTM.

nuttycom (dismissed):
utACK with minor naming nit

src/interpreter.rs
```diff
@@ -460,26 +460,58 @@ const VCH_TRUE: [u8; 1] = [1];
...
+        Self::initial(Stack(vec![]))
+    }
+
+    /// Creates a state with an initial stack, but other components empty.
     pub fn initial(stack: Stack<Vec<u8>>) -> Self {
+        Self::manual(stack, Stack(vec![]), 0, Stack(vec![]))
+    }
+
+    /// Create an arbitrary stat.
+    pub fn manual(
```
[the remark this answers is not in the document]
sellout: I renamed it, but as least as it currently is, it’s not only used for testing – it’s used in the definition of `State::initial` (and transitively `State::new`), so `from_parts` is the only function that directly constructs a `State`.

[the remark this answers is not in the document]
nuttycom: In this circumstance, I would use the private constructor directly for `new` and `initial`, and then only expose `from_parts` under the feature flag. If there's not a good reason for it to explicitly be part of the public API, let's keep it private so that it's not a potential footgun.

nuttycom (approved):
re-utACK https://github.com/ZcashFoundation/zcash_script/commit/b3ba41dddb7ec7796a26dfc1159c9be51b8dc9fa

sellout:
> This looks great, but I'd like to understand the big picture - is it worth it improving the testability and debuggability of what is probably the most ossified part of the codebase? Is there further work planned on top of this?

Yes – this is in support of the Rust implementation, which has ongoing changes (e.g., sellout/zcash_script#1 and sellout/zcash_script#2). So this is to help ensure that we know exactly when and where it diverges from the C++ behavior moving forward, especially as new functionality (like that to support PCZTs) is integrated.

PR #208, opened 2025-04-15
Eliminate `ScriptNum`

sellout:
__NB__: I recommend reviewing commit-by-commit, and ignoring whitespace when reviewing the second commit.

`ScriptNum` was used one of three ways:
1. convert a Rust integer to a stack value,
2. load a stack value into Rust, xor
3. arithmetic on stack values.

Combining these into one interface added extra complexity. For example, `getint()` would clamp the value to `i32` range, but that range could only be violated if arithmetic was performed on `ScriptNum`s, and `getint()` was never used on the result of arithmetic.

This extracts `parse_num` and `serialize_num` and adds a number of small functions that encapsulate the different arithemetic operations. The above patterns have changed in the following ways:
1. `ScriptNum::from(_).getvch()` is now `serialize_num(_)`, and
2. `ScriptNum::new(_).and_then(_.getint())` is now `parse_num(_)`,
3. `ScriptNum::new(_)` … `getvch()` uses `binop(_, _)` and friends.

Because only `i32` values are read from the stack, but `i64` results are pushed back onto the stack[^1], the arithmetic operations that exist can’t possibly hit the various bounds that were checked on `ScriptNum` operations, so they were removed.

[^1]: This does mean that a script like `[0, 1, i32::MAX, OP_ADD, OP_ADD]` will fail with `ScriptNumError::Overflow`[^2] when the second `OP_ADD` tries to read `i32::MAX+1` as its second argument, but them’s the semantics we were given.

[^2]: Even worse, in the C++ impl, this simply comes back as `UnknownError`.

As part of the above work, this introduces `cast_from_bool`, which is then used to replace all instances of `VCH_TRUE`/`FALSE`.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1,4 +1,7 @@
-use std::slice::Iter;
+use std::{
+    cmp::{max, min},
+    slice::Iter,
+};
 
 use ripemd::Ripemd160;
 use sha1::Sha1;
@@ -461,8 +464,50 @@ fn check_minimal_push(data: &[u8], opcode: PushValue) -> bool {
     true
 }
 
-const VCH_FALSE: ValType = Vec::new();
-const VCH_TRUE: [u8; 1] = [1];
+fn unop<T: Clone>(
+    stack: &mut Stack<T>,
+    op: impl Fn(T) -> Result<T, ScriptError>,
+) -> Result<(), ScriptError> {
+    let item = stack.pop()?;
+    op(item).map(|res| stack.push_back(res))
+}
+
+fn binfn<T: Clone, R>(
+    stack: &mut Stack<T>,
+    op: impl Fn(T, T) -> Result<R, ScriptError>,
+) -> Result<R, ScriptError> {
+    let x2 = stack.pop()?;
+    let x1 = stack.pop()?;
+    op(x1, x2)
+}
+
+fn binbasic_num<R>(
+    stack: &mut Stack<Vec<u8>>,
+    require_minimal: bool,
+    op: impl Fn(i64, i64) -> Result<R, ScriptError>,
+) -> Result<R, ScriptError> {
+    binfn(stack, |x1, x2| {
+        let bn2 = parse_num(&x2, require_minimal, None).map_err(ScriptError::ScriptNumError)?;
+        let bn1 = parse_num(&x1, require_minimal, None).map_err(ScriptError::ScriptNumError)?;
+        op(bn1, bn2)
+    })
+}
+fn binop<T: Clone>(
+    stack: &mut Stack<T>,
+    op: impl Fn(T, T) -> Result<T, ScriptError>,
+) -> Result<(), ScriptError> {
+    binfn(stack, op).map(|res| stack.push_back(res))
+}
+
+fn cast_from_bool(b: bool) -> ValType {
+    static VCH_FALSE: [u8; 0] = [];
+    static VCH_TRUE: [u8; 1] = [1];
+    if b {
+        VCH_TRUE.to_vec()
+    } else {
+        VCH_FALSE.to_vec()
+    }
+}
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct State {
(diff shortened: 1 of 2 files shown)


src/script.rs
```diff
@@ -265,219 +260,113 @@ impl From<Operation> for u8 {
...
-                .expect("caller should avoid overflow"),
-        )
+    let mut result = Vec::new();
+    let neg = value < 0;
+    let mut absvalue = value.abs();
+
+    while absvalue != 0 {
+        result.push(
+            (absvalue & 0xff)
+                .try_into()
```
nuttycom, line 348: nonblocking nit: recent experiences with compilation failures resulting from ambiguous `From` impls being introduced at different points in a dependency stack have made me prefer the more explicit `u8::try_from` to `try_into` - I basically no longer use `.into` or `.try_into` anywhere; it makes it easier to see the type and prevents spooky action at a distance across a dependency stack.

nuttycom (approved):
utACK 2d52528e90b3e5ff009b4a1e0ffad5128b1d467e

src/interpreter.rs
```diff
@@ -642,18 +722,17 @@ pub fn eval_step<'a>(
                             // Thus as a special case we tell `ScriptNum` to accept up
```
daira, line 722:
```suggestion
                            // Thus as a special case we tell `parse_num` to accept up
```

src/interpreter.rs
```diff
@@ -912,139 +989,70 @@ pub fn eval_step<'a>(
...
-                        stack.pop()?;
-                        stack.push_back(if equal { VCH_TRUE.to_vec() } else { VCH_FALSE });
-                        if op == OP_EQUALVERIFY {
-                            if equal {
-                                stack.pop()?;
-                            } else {
-                                return set_error(ScriptError::EqualVerify);
-                            }
+                    // (x1 x2 - bool)
+                    OP_EQUAL => binop(stack, |x1, x2| Ok(cast_from_bool(x1 == x2)))?,
```
daira, line 1001: This drops the explicit check for `InvalidStackOperation`. If I understand correctly, it's correct to do that here because `stack.top` will do it, and there are no other side effects / early returns before that could occur.

But there are many many other places where that is the case (and the explicit checks each take 3 lines so they are adding to clutter). Is the plan to remove those soon? If so then no problem, it doesn't need to be in this PR.

src/script.rs
```diff
@@ -265,219 +260,113 @@ impl From<Operation> for u8 {
...
+                .unwrap_or_else(|_| unreachable!()),
+        );
```
daira, line 350: I agree with @nuttycom. In this case it helps to see that the out-of-range cannot occur. Also prefer `expect` to `.unwrap_or_else(|_| unreachable!())` because the message to "expect" can help to clarify why there cannot be a panic.

```suggestion
        result.push(u8::try_from(absvalue & 0xff).expect("fits in u8"));
```

daira (approved):
utACK with comments.

src/script.rs
nuttycom: Resolving as a nit so that this can merge.

src/interpreter.rs
nuttycom: Resolving as a nit so that this can merge.

src/interpreter.rs
```diff
@@ -535,6 +582,39 @@ pub fn eval_step<'a>(
...
+        };
+
+    let binrel =
+        |stack: &mut Stack<Vec<u8>>, op: &dyn Fn(i64, i64) -> bool| -> Result<(), ScriptError> {
+            binfn_num(stack, &|bn1, bn2| cast_from_bool(op(bn1, bn2)))
+        };
+
+    let unrel = |stack: &mut Stack<Vec<u8>>, op: &dyn Fn(i64) -> bool| -> Result<(), ScriptError> {
+        unfn_num(stack, &|bn| cast_from_bool(op(bn)))
+    };
```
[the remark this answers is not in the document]
nuttycom: Resolving as a nit so that this can merge.

PR #209, opened 2025-04-15
Stable Rust changes

sellout:
This contains a number of changes to the Rust implementation that maintain stepwise identical behavior with the C++ implementation.

It should be reviewed commit-by-commit.

### DIFF
--- a/fuzz/fuzz_targets/compare.rs
+++ b/fuzz/fuzz_targets/compare.rs
@@ -6,12 +6,12 @@ extern crate zcash_script;
 use zcash_script::{
     check_verify_callback,
     interpreter::{CallbackTransactionSignatureChecker, VerificationFlags},
-    normalize_err, rust_interpreter,
+    normalize_err, rust_interpreter, script,
     signature::HashType,
     testing, CxxInterpreter,
 };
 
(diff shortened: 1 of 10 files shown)


sellout:
This PR can potentially have some parts rejected, so if any of the commits (or smaller parts) seem unnecessary, we can discuss. I don’t always do a good job of including justification in the commit messages.

E.g., the change to have `PushValue`s carry their data simplifies representation of the script as `Vec<Opcode>`, which is useful in cases like error reporting (as we can include the _script_ as opposed to opaque bytes) and PCZTs (which can accept a `Vec<Opcode>`, which is then serialized into the tx).

nuttycom (commented):
Flushing comments; I still need to spend some more time with 05ba17bb064682fbb8b1f71cfd8db62c289150cd to make sure I've checked everything.

sellout:
Sorry, there has been a lot of churn here after addressing @nuttycom’s comments. Lesson learned – this PR is just too big. I ended up removing the `no_std` changes, because they weren’t as trivial as I initially thought.

Everything is passing now, though, so should be good.

sellout:
I am splitting this PR into multiple smaller ones. I’ll update this comment with the list of new PRs.

* #221

PR #210, opened 2025-04-21
C++ stepwise comparison

sellout:
This PR exposes stepwise evaluation on the C++ and runs it against the Rust side to ensure the state is identical every step of the way.

There are two commits here. The first one makes minimal changes to the C++ interpreter to expose an `EvalStep` function. It should be reviewed with whitespace ignored, since the body of the interpreter is reindented without changes.

The second commit implements the comparison work, changing the `zcash_script.h` interface.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,7 @@ impl std::error::Error for Error {}
 // to pull a newer version, so this silences the deprecation on that job.
 #[allow(deprecated)]
 fn bindgen_headers() -> Result<()> {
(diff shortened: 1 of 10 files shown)


sellout:
BLOCKER: Stepwise comparison currently leaks `ZcashScriptState` on each call to `zcash_script_eval_step`. Before I spend too much time on it, I’d like to know whether the FFI approach I’m using is right and what the best way to delete these vectors (actually, pointers to `vector::data()`) would be.

PR #216, opened 2025-05-15
Remove extra `&` from `SighashCalculator`

sellout:
It‘s odd that the previous code worked (in some cases).


conradoplg (approved):
Thanks!

PR #217, opened 2025-05-17
bump to 0.3.1

conradoplg:
also sneaked in the dependabot change to only updated lockfile, and some related Cargo.toml adjustments

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,9 @@ version: 2
 updates:
   - package-ecosystem: cargo
     directory: '/'
+    # Update only the lockfile. We shouldn't update Cargo.toml unless it's for
+    # a security issue, or if we need a new feature of the dependency.
+    versioning-strategy: lockfile-only
     schedule:
       interval: daily
       timezone: America/New_York
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zcash_script"
-version = "0.3.0"
+version = "0.3.1"
 authors = [
     "Electric Coin Company <info@electriccoin.co>",
     "Greg Pfeil <greg@technomadic.org>",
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -73,25 +73,21 @@ ripemd = "0.1"
...
-cc = { version = "1.2.11", features = ["parallel"] }
+# We want zcash-script to be compatible with whatever version of `bindgen` that
+# is being indirectly used by Zebra via the `rocksdb` dependency. To avoid
+# having to update zcash-script every time Zebra updates its `rocksdb`
+# dependency, we allow any version of `bindgen`, even major version bumps, by
+# using the `>=` operator. There is some chance that this breaks the build if a
+# breaking API change affects us, but we can always fix it then.
+bindgen = ">= 0.69.5"
+# Same reasoning as above
+cc = { version = ">= 1.2.11", features = ["parallel"] }
```
conradoplg, line 87: We used to have `>=` here but it was lost in some dependabot upgrade. I added it back and changed the wording of the comment above to be hopefully more clear about the reasoning.

The bindgen bump was required because this really does not compile with `0.64.0` anymore. (Tested with `cargo +nightly update -Z direct-minimal-versions`)

Cargo.toml
```diff
@@ -73,25 +73,21 @@ ripemd = "0.1"
 secp256k1 = "0.30"
 sha-1 = "0.10"
 sha2 = "0.10"
-tracing = "0.1.41"
```
conradoplg, line 76: Bumping this down to not add tight version restrictions to crate users. I tested that 0.1.0 works by using `cargo +nightly update -Z direct-minimal-versions` (we can add that test to CI later)

.github/dependabot.yml
```diff
@@ -2,6 +2,9 @@ version: 2
 updates:
   - package-ecosystem: cargo
     directory: '/'
+    # Update only the lockfile. We shouldn't update Cargo.toml unless it's for
+    # a security issue, or if we need a new feature of the dependency.
+    versioning-strategy: lockfile-only
```
conradoplg, line 7: See https://github.com/ZcashFoundation/zebra/issues/9398 for context

upbqdn (approved):

PR #220, opened 2025-05-21
Update the Rust toolchain

sellout:
It was pinned to 1.81 because of a C++ linking issue in later versions. But now that 1.85 (and later) has been released, we can move to “stable”, and bump the MSRV (because it was previously the MaximumRSV, since later releases didn’t work).

This also makes bindgen track the MSRV more consistently for the code it generates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [package]
+rust-version = "1.85" # C++ linking was broken in rust 1.82–1.84
 name = "zcash_script"
 version = "0.3.2"
 authors = [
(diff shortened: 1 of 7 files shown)


sellout:
Hrmm, when I wrote https://github.com/ZcashFoundation/zcash_script/pull/176#issuecomment-2524395651, I assumed that what was “nightly” at the time would become 1.85, but now 1.85 (and 1.87) have been released, and they still exhibit the same linking issue, while the currently nightly continues to work.

arya2:
> I assumed that what was “nightly” at the time would become 1.85, but now 1.85 (and 1.87) have been released, and they still exhibit the same linking issue, while the currently nightly continues to work.

There are features/fixes on nightly that became available several years ago and still aren't available on stable/beta.

PR #221, opened 2025-05-22
Simplify handling of ECDSA signatures

sellout:
The C++ implementation had a lot of manual checking of bytes. In Rust, we have the secp256k1 crate, which takes care of most of this.

### DIFF
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -247,10 +247,13 @@ fn is_sig_valid(
         signature::Validity::InvalidAbort(e) => Err(e.into()),
         signature::Validity::InvalidContinue => {
             // We still need to check the pubkey here, because it can cause an abort.
-            check_pub_key_encoding(vch_pub_key, flags).map(|()| false)
+            check_pub_key_encoding(vch_pub_key, flags)?;
+            Ok(false)
         }
-        signature::Validity::Valid(sig) => check_pub_key_encoding(vch_pub_key, flags)
-            .map(|()| checker.check_sig(&sig, vch_pub_key, script)),
+        signature::Validity::Valid(sig) => {
+            check_pub_key_encoding(vch_pub_key, flags)?;
+            Ok(checker.check_sig(&sig, vch_pub_key, script))
+        }
     }
 }
 
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -108,8 +108,6 @@ pub enum SignedOutputs {
 /// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct HashType {
-    /// This is in `Option`, because if `VerificationFlags::StrictEnc` isn’t set, a value that
-    /// doesn’t correspond to any signed outputs setting is possible.
     signed_outputs: SignedOutputs,
     anyone_can_pay: bool,
 }


sellout:
This PR has been extracted from #209.

src/external/pubkey.rs
```diff
@@ -1,56 +1,37 @@
...
         if let Ok(pubkey) = PublicKey::from_slice(self.0) {
-            // let sig: secp256k1_ecdsa_signature;
-            if vch_sig.is_empty() {
-                return false;
-            };
-            // Zcash, unlike Bitcoin, has always enforced strict DER signatures.
-            if let Ok(mut sig) = ecdsa::Signature::from_der(vch_sig) {
-                // libsecp256k1's ECDSA verification requires lower-S signatures, which have
-                // not historically been enforced in Bitcoin or Zcash, so normalize them first.
-                sig.normalize_s();
```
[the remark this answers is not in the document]
sellout: I should have commented on this in the commit message. In

https://github.com/ZcashFoundation/zcash_script/blob/1d86418b2c0594e2f6a2173a0eb1b47919751e1f/src/interpreter.rs#L358-L362

`check_signature_encoding` calls `check_low_s` (defined just below here), and that results in a `SigHighS` failure if the signature doesn’t already match the result of `normalize_s`. So when `check_sig` calls `verify`, we already know we have a normalized `s`.

The motivation for the change was to avoid either copying or making the signature `mut`.

conradoplg:
I'll merge this soon unless @str4d wants to take a look as well?

conradoplg:
This ended up having conflicts with the just-merged https://github.com/ZcashFoundation/zcash_script/pull/231, let me know if you would prefer for me to sort them out

src/script_error.rs
```diff
@@ -1,3 +1,15 @@
+use secp256k1;
+
+/// Things that can go wrong when constructing a `HashType` from bit flags.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum InvalidHashType {
```
str4d, line 5: Document this addition in the changelog.

src/script_error.rs
```diff
@@ -39,8 +51,8 @@ pub enum ScriptError {
     UnsatisfiedLockTime,
 
     // BIP62
-    SigHashType,
-    SigDER,
+    SigHashType(Option<InvalidHashType>),
+    SigDER(Option<secp256k1::Error>),
```
str4d, line 55: Document this change to the enum in the changelog.

src/script_error.rs
```diff
@@ -1,5 +1,16 @@
+use secp256k1;
 use thiserror::Error;
 
+/// Things that can go wrong when constructing a `HashType` from bit flags.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum InvalidHashType {
+    /// Either or both of the two least-significant bits must be set.
+    UnknownSignedOutputs,
```
str4d, line 8: Check whether these need to have `thiserror` derives (per #231).

src/interpreter.rs
```diff
@@ -260,179 +257,52 @@ impl<T: Clone> Stack<T> {
...
+    sighash: HashType,
 }
 
-fn is_defined_hashtype_signature(vch_sig: &ValType) -> bool {
-    if vch_sig.is_empty() {
-        return false;
-    };
-
-    HashType::from_bits(i32::from(vch_sig[vch_sig.len() - 1]), true).is_ok()
+fn decode_signature(vch_sig_in: &[u8], is_strict: bool) -> Result<Option<Signature>, ScriptError> {
```
str4d, line 274: This function needs documentation. The fact that `Ok(None)` is effectively a deferred error is not at all clear, and now that `decode_signature` has a return value (vs `check_signature_encoding`), it is important that the caller handle it correctly (which they may not do if they don't go look at the internals of this method each time they interact with its callsites).

src/interpreter.rs
```diff
@@ -260,179 +257,52 @@ impl<T: Clone> Stack<T> {
...
+    match vch_sig_in.split_last() {
+        // Empty signature. Not strictly DER encoded, but allowed to provide a compact way to
+        // provide an invalid signature for use with CHECK(MULTI)SIG
+        None => Ok(None),
+        Some((hash_type, vch_sig)) => Ok(Some(Signature {
+            sig: ecdsa::Signature::from_der(vch_sig).map_err(|e| ScriptError::SigDER(Some(e)))?,
+            sighash: HashType::from_bits((*hash_type).into(), is_strict)
+                .map_err(|e| ScriptError::SigHashType(Some(e)))?,
```
str4d, line 282: Consensus bug: in the code prior to this PR, a `HashType::from_bits` error would result in `false` being pushed to the stack (via `CallbackTransactionSignatureChecker::check_sig`). Now it results in an error being raised, halting stack execution.

src/interpreter.rs
```diff
@@ -260,179 +257,52 @@ impl<T: Clone> Stack<T> {
...
-    };
-
-    HashType::from_bits(i32::from(vch_sig[vch_sig.len() - 1]), true).is_ok()
+fn decode_signature(vch_sig_in: &[u8], is_strict: bool) -> Result<Option<Signature>, ScriptError> {
+    match vch_sig_in.split_last() {
+        // Empty signature. Not strictly DER encoded, but allowed to provide a compact way to
+        // provide an invalid signature for use with CHECK(MULTI)SIG
+        None => Ok(None),
+        Some((hash_type, vch_sig)) => Ok(Some(Signature {
+            sig: ecdsa::Signature::from_der(vch_sig).map_err(|e| ScriptError::SigDER(Some(e)))?,
```
str4d, line 280: Consensus bug: in the code prior to this PR, a ` ecdsa::Signature::from_der` error would result in false being pushed to the stack (via `Pubkey::verify`). Now it results in an error being raised, halting stack execution.

str4d (changes requested):
Flushing partial review as of a46ae1b4c6ea5e2dcae47fd7cf0470057fa0f94b.

src/signature.rs
```diff
@@ -0,0 +1,202 @@
...
+                            .map_err(|e| Error::SigDER(Some(e)))
+                            .and_then(|sig| {
+                                if require_low_s && !PubKey::check_low_s(&sig) {
+                                    Err(Error::SigHighS)
+                                } else {
+                                    Ok(sig)
+                                }
+                            })?,
+                        sighash: HashType::from_bits((*hash_type).into(), is_strict)
+                            .map_err(|e| Error::SigHashType(Some(e)))?,
```
str4d, line 186: Consensus bug: in the code prior to this PR, `HasType::from_bits` was called in `CallbackTransactionSignatureChecker::check_sig`, and a parse failure was mapped to `false`, which results in the stack continuing to be processed (either pushing `false` to the stack, or trying the signature against the next pubkey). Now it results in an error being raised, halting stack execution.

(This is https://github.com/ZcashFoundation/zcash_script/pull/221#discussion_r2164620879 which was unaddressed, but the code moved to a different location.)

The fact that the now-imported test vectors pass indicates that the specific test about signature/pubkey validation ordering for `CHECKMULTISIG` does not cover an invalid signature due to invalid sighash type. We should add a test vector that fails for that reason, and confirm it passes before this PR.

---

While reviewing the `ecdsa::Signature::from_der` usage, I realised that the issue here is that `HashType::from_bits` is now doing two different things:
- It is replacing the `is_low_der_signature` call in `check_signature_encoding`, which _does_ raise an error and halt stack execution, but only when `flags.contains(VerificationFlags::StrictEnc)`.
- It is replacing its previous call in `CallbackTransactionSignatureChecker::check_sig` (which was previously called after `check_signature_encoding`), which _does not_ halt stack execution.

So AFAICT, whether or not an error should be raised depends on `is_strict`. If `Decoded::from_bytes` were not being added to the public API, then the only place it would be used is inside `is_sig_valid`, and the way it is used there means that I think the "stack execution does not halt on invalid" case could be handled by returning `Ok(None)` (as is done for an empty signature). But if `Decoded::from_bytes` stays in the public API, then either its documentation needs to change to explain what `Ok(None)` means now, or we need a wrapper around it that maps the specific error we care about here to `Ok(None)` when `flags.contains(VerificationFlags::StrictEnc)`.

(I also just noticed that the old `is_defined_hashtype_signature` resulted in an error being raised for an empty signature, so we should _also_ check that behaviour, because that might also have a consensus bug in the opposite StrictEnc setting from this one.)

src/signature.rs
```diff
@@ -0,0 +1,202 @@
...
+        require_low_s: bool,
+        is_strict: bool,
+    ) -> Result<Option<Self>, Error> {
+        match vch_sig_in.split_last() {
+            None => Ok(None),
+            Some((hash_type, vch_sig)) => {
+                if Self::is_valid_encoding(vch_sig) {
+                    Ok(Some(Decoded {
+                        sig: ecdsa::Signature::from_der(vch_sig)
+                            .map_err(|e| Error::SigDER(Some(e)))
```
str4d, line 177: Possible consensus bug: this is unaltered from https://github.com/ZcashFoundation/zcash_script/pull/221#discussion_r2164623190, but now that various checks in `is_valid_signature_encoding` have been removed due to duplicating checks in `ecdsa::Signature::from_der`, I think that this error being raised might be fine (because prior to this PR, `is_valid_signature_encoding` failing results in script execution being aborted).

However, if `ecdsa::Signature::from_der` performs _any_ checks that were previously not performed in `is_valid_signature_encoding`, then the previous comment applies and there is a consensus bug here (because prior to this PR, those checks resulted in script execution continuing).

Additionally, even if `ecdsa::Signature::from_der` currently performs precisely the same checks as were removed from `is_valid_signature_encoding`, the refactor means that the behaviour of `ecdsa::Signature::from_der` is consensus-critical. Document this criticality in `Cargo.toml` beside the `secp256k1` crate version definition, so that when we perform updates we know to review the update for consensus compatibility.

str4d (changes requested):
Reviewed f246e472e34346985ee84b8eafb333ba1f77aab5.

src/external/pubkey.rs
```diff
@@ -1,56 +1,41 @@
...
 impl PubKey<'_> {
-    pub const PUBLIC_KEY_SIZE: usize = 65;
-    pub const COMPRESSED_PUBLIC_KEY_SIZE: usize = 33;
+    pub const SIZE: usize = 65;
+    pub const COMPRESSED_SIZE: usize = 33;
 
     /// Check syntactic correctness.
     ///
-    /// Note that this is consensus critical as CheckSig() calls it!
+    /// Note that this is consensus critical as `check_sig` calls it!
```
daira, line 11:
```suggestion
    /// Note that this is consensus-critical as `check_sig` calls it!
```
Non-blocking.

src/external/pubkey.rs
```diff
@@ -1,56 +1,41 @@
...
-                sig.normalize_s();
-                let secp = Secp256k1::verification_only();
-                secp.verify_ecdsa(&Message::from_digest(*hash), &sig, &pubkey)
-                    .is_ok()
-            } else {
-                false
-            }
+            let mut normalized_sig = *sig;
+            // libsecp256k1's ECDSA verification requires lower-S signatures, which have
+            // not historically been enforced in Bitcoin or Zcash, so normalize them first.
```
daira, line 26:
```suggestion
            // libsecp256k1's ECDSA verification requires lower-S signatures, which are
            // not required by consensus in Zcash, so normalize them first.
```

src/interpreter.rs
```diff
@@ -455,6 +230,27 @@ fn check_minimal_push(data: &[u8], opcode: PushValue) -> bool {
...
 
+fn is_sig_valid(
+    vch_sig: &[u8],
+    vch_pub_key: &[u8],
+    flags: VerificationFlags,
+    script: &Script<'_>,
+    checker: &dyn SignatureChecker,
+) -> Result<bool, ScriptError> {
+    // Note how this makes the exact order of pubkey/signature evaluation distinguishable by
+    // CHECKMULTISIG NOT if the STRICTENC flag is set. See the script_(in)valid tests for details.
```
daira, line 241: At some point I want to simplify the whole verification flags thing. In order of increasing possible controversy:
* we always check the things that are Zcash consensus rules (we did this for SCRIPT_VERIFY_DERSIG but we did not do it for SCRIPT_VERIFY_P2SH);
* we strip out the code for SCRIPT_VERIFY_SIGPUSHONLY which has never been set in Zcash (`IsPushOnly` is in any case verified for P2SH scripts);
* we just have one flag for whether we are also checking the other rules (SCRIPT_VERIFY_{STRICTENC, MINIMALDATA, NULLDUMMY, DISCOURAGE_UPGRADABLE_NOPS, CLEANSTACK, CHECKLOCKTIMEVERIFY, LOW_S});
* in some transaction version update (not NU) where we are not doing too much else, we make those also consensus rules for that tx version and later.

This would allow simplifications from not having to consider combinations of flags that are never used and never will be used in Zcash. It would also be less of a compatibility burden for the ecosystem to do it in a tx version rather than an NU (since they can always just not generate that tx version for now if it causes them problems). That tx version will eventually be forbidden and then we get a further simplification.

src/signature.rs
```diff
@@ -0,0 +1,202 @@
...
+    anyone_can_pay: bool,
+}
+
+impl HashType {
+    /// Construct a `HashType` from bit flags.
+    ///
+    /// ## Consensus rules
+    ///
+    /// [§4.10](https://zips.z.cash/protocol/protocol.pdf#sighash):
+    /// - Any `HashType` in a v5 transaction must have no undefined bits set.
```
daira, line 79:
```suggestion
    /// - Any `SighashType` in a v5 transaction must have no undefined bits set.
```

src/signature.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub fn from_bits(bits: i32, is_strict: bool) -> Result<Self, InvalidHashType> {
+        let unknown_bits = (bits | 0x83) ^ 0x83;
+        if is_strict && unknown_bits != 0 {
+            Err(InvalidHashType::ExtraBitsSet(unknown_bits))
+        } else {
+            let msigned_outputs = match (bits & 2 != 0, bits & 1 != 0) {
+                (false, false) => Err(InvalidHashType::UnknownSignedOutputs),
```
daira, line 86:
```suggestion
                (false, false) => Err(InvalidSighashType::UnknownSignedOutputs),
```

src/signature.rs
```diff
@@ -0,0 +1,202 @@
...
+            }
+        }
+    }
+
+    pub fn sig(&self) -> &ecdsa::Signature {
+        &self.sig
+    }
+
+    pub fn sighash(&self) -> &HashType {
+        &self.sighash
```
daira, line 200:
```suggestion
    pub fn sighash_type(&self) -> &SigashType {
        &self.sighash_type
```

src/interpreter.rs
```diff
@@ -1283,30 +1073,22 @@ pub const SIGHASH_SIZE: usize = 32;
...
-        match vch_sig_in.split_last() {
-            None => false,
-            Some((hash_type, vch_sig)) => HashType::from_bits((*hash_type).into(), false)
-                .ok()
-                .and_then(|hash_type| (self.sighash)(script_code.0, hash_type))
-                .map(|sighash| Self::verify_signature(vch_sig, &pubkey, &sighash))
-                .unwrap_or(false),
-        }
+        pubkey.is_valid()
+            && (self.sighash)(script_code.0, sig.sighash())
```
daira, line 1089:
```suggestion
            && (self.sighash)(script_code.0, sig.sighash_type())
```

CHANGELOG.md
```diff
@@ -8,6 +8,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `interpreter::CallbackTransactionSignatureChecker::verify_signature` was an internal function that was accidentally exposed
+
+### Changed
+
+- `HashType`, `InvalidHashType`, and `SignedOutputs` have moved from `interpreter` to `signature`
+- the `SigHashType`, `SigDER`, and `SigHighS` `ScriptError` variants have been moved to the new `signature::Error`, and are grouped under the new `ScriptError::SignatureEncoding`
+- `interpreter::SighashCalculator` now takes `&HashType` instead of `HashType`
+- `interpreter::Signature` is now `signature::Decoded`
+- `interpreter::SignatureChecker::check_sig` takes a `&signature::Decoded` instead of a `&[u8]`
+- `signature::HashType` now has private fields, but there are now functions with the same names for reading
```
daira, line 24:
```suggestion
- `signature::HashType` now has private fields, but there are corresponding getters
```

src/signature.rs
sellout: This has now been changed to effectively result in `false`, and the former `is_valid_[signature]_encoding` is restored to its former power.

nuttycom (approved):
utACK 27ef4fc158f00f6652b4ae025c5dd25aeaab39dd

src/interpreter.rs
nuttycom: Recorded as #249

nuttycom (approved):
re-utACK 9257929ace52ecad80b6a696901e5132f89188dd

str4d (approved):
utACK 9257929ace52ecad80b6a696901e5132f89188dd

src/signature.rs
```diff
@@ -116,15 +158,26 @@ pub struct Decoded {
...
+            [] => Err(InvalidDerInteger::ZeroLength),
             // Null bytes at the start are not allowed, unless it would otherwise be interpreted as
             // a negative number.
-            [0x00, next, ..] => next & 0x80 != 0,
+            [0x00, next, ..] => {
+                if next & 0x80 != 0 {
+                    Ok(())
+                } else {
+                    Err(InvalidDerInteger::LeadingNullByte)
+                }
```
daira, line 185: This could be
```rust
(next & 0x80 != 0).then_some(()).ok_or(InvalidDerInteger::LeadingNullByte)
```
I'm not sure whether that's better. :woman_shrugging:

src/signature.rs
```diff
@@ -157,27 +210,65 @@ impl Decoded {
...
+                    // Make sure the length covers the entire signature.
+                    if usize::from(*total_len) == content.len() {
+                        match content {
+                            // Check whether the R element is an integer.
+                            // Extract the length of the R element.
+                            [0x02, r_len, r_s @ ..] => match r_s.split_at((*r_len).into()) {
+                                // Check whether the S element is an integer.
+                                // Extract the length of the S element.
+                                // Make sure the length of the S element is still inside the signature.
+                                (r, [0x02, s_len, s @ ..]) => Self::is_valid_integer(r)
```
daira, line 237: If you pass the name and length into `Self.is_valid_integer` (maybe call it `parse_integer`), then it can return a `Result` itself, and I think in that case you can deduplicate the error handling between `r` and `s`. (The length of `r_s` will be checked redundantly but it doesn't matter.)

src/signature.rs
```diff
@@ -108,7 +108,9 @@ pub enum SignedOutputs {
 /// The different SigHash types, as defined in <https://zips.z.cash/zip-0143>
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct HashType {
-    signed_outputs: SignedOutputs,
+    /// This is in `Option`, because if `VerificationFlags::StrictEnc` isn’t set, a value that
```
daira: "an `Option`"

daira (commented):
Post-hoc utACK

src/interpreter.rs
```diff
@@ -251,179 +190,15 @@ impl<T: Clone> Stack<T> {
...
 }
 
-fn is_compressed_or_uncompressed_pub_key(vch_pub_key: &ValType) -> bool {
-    if vch_pub_key.len() < PubKey::COMPRESSED_PUBLIC_KEY_SIZE {
-        //  Non-canonical public key: too short
-        return false;
+fn is_compressed_or_uncompressed_pub_key(vch_pub_key: &[u8]) -> bool {
+    match vch_pub_key.first() {
+        Some(0x02 | 0x03) => vch_pub_key.len() == PubKey::COMPRESSED_SIZE,
+        Some(0x04) => vch_pub_key.len() == PubKey::SIZE,
```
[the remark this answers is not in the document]
daira: It isn't too restrictive; it matches zcashd: https://github.com/zcash/zcash/blob/fa9d32b3e0ad7920d11ddcfec4fbfd9de4c99f7e/src/script/interpreter.cpp#L72-L92

PR #231, opened 2025-06-16
add: Implement `Display` and derive `Error` traits for error enums

upbqdn:
## Motivation

I needed the [`Error`](https://github.com/zcashfoundation/zcash_script/blob/87a3f4b7b23bd232128bd65cf10ac1a43e90a543/src/zcash_script.rs#L13) enum to implement the [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) trait so I could use it in Zebra conveniently.

## Solution

- Implement `Display` and derive [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) for all relevant types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,9 +587,9 @@ checksum = "0fda2ff0d084019ba4d7c6f371c95d8fd75ce3524c3cb8fb653a3023f6323e64"
 
 [[package]]
 name = "syn"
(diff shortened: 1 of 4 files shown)


upbqdn:
Thanks for the suggestion. I didn't think of it originally. I implemented it.

conradoplg (approved):

PR #234, opened 2025-06-23
downgrade secp256k1 back to 0.29

### LINKED ISSUE #232: Downgrade to `secp256k1 0.29`
#191 was merged in February, and then included in the 0.3.0 release alongside the initial Rust changes. This introduced two versions of `secp256k1` into the Zebra consensus code. I believe this was not noticed because the new Rust code is consumed through the C++ interface, so everything is serialized and then unserialized.

More of a problem is that the `zcash_transparent` crate is currently requires to use `secp256k1 0.29` because that is the version that the Keystone firmware uses. We therefore cannot use `zcash_script` inside `zcash_transparent` unless `secp256k1` is downgraded.

Looking over the current codebase with @sellout, @nuttycom, and @daira, it appears that `secp256k1` is not used anywhere in the public API currently. So we can currently perform the downgrade in a patch release. In future it is likely that the Rust API will start to expose `secp256k1` semantics, but for now we are lucky.

conradoplg:
Closes #232

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,12 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d5a26814d8dcb93b0e5a0ff3c6d80a8843bafb21b39e8e18a6f05471870e110"
 
 [[package]]
-name = "arrayvec"
-version = "0.7.6"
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK f1fb7345a1898624d7c07bd423610b3927a9f6d1

Issue #240, opened 2025-07-02
some tests need correct sighash function


sellout:
When #157 modifed the C++ implementation to use a callback, script_tests.cpp wasn’t updated, so it no longer compiles. Those tests are ported over to Rust as `TestVector`s, but some of them fail (agaist both the pure Rust and wrapped C++ implementations) because they need a sighash function that replicates [the pre-callback behavior](https://github.com/ZcashFoundation/zcash_script/blob/fa1bb2571dc95d59d6864f738a39b461d666b57a/depend/zcash/src/test/script_tests.cpp#L44-L97), however the point of the callback is to remove dependencies on other zcash crates, and so this one can no longer trivially construct `Transaction`s, etc.

We should be able to construct one that uses a mostly-hardcoded serialized transaction that we can insert the serialized script components into and then hash, then tests that depend on the `CHECK*SIG*` opcodes behaving correctly can work.

PR #241, opened 2025-07-02
Migrate script_*valid.json tests to Rust `TestVector`s

sellout:
This brings both versions of the tests (JSON[^1] & TestVector) up-to-date with [Bitcoin](https://github.com/bitcoin/bitcoin/commit/fa33592898838cf92ae3bdb3ca9a522b5f9cce3d) (modulo the tests for features/flags Zcash doesn’t support).

[^1]: The C++ implementation can’t run the JSON tests since the [switch to callbacks](#157) anyway, so I allowed the format to update with Bitcoin.

This also pulls in some of ergonomic opcode work from #223, since the TestVectors really benefit from that brevity.

However, some tests currently have incorrect expectations (tagged with references to #240), this is because they rely on specific transactions in `CHECK*SIG*`, which is harder to support after the change to callbacks.


conradoplg (approved):
Looks good, thanks

PR #245, opened 2025-07-11
ci: remove cargo dependabot

conradoplg:
Dependabot for Rust it not generally useful for this repo. We should update dependencies manually if required before each release.


natalieesk (approved):
Looks good

PR #247, opened 2025-07-24
Add `standard::solver` function for identifying standard scripts

str4d:
Ported from `zcashd`.

Only the last commit is relevant; it is based on the rest of @sellout's in-progress changes.


src/standard.rs
```diff
@@ -0,0 +1,93 @@
...
+///
+/// Returns `None` if the script is non-standard.
+pub fn solver(script_code: &script::PubKey) -> Option<ScriptKind> {
+    match &script_code.0[..] {
+        // Pay-to-Script-Hash (P2SH)
+        [op::HASH160, Opcode::PushValue(PushValue::LargeValue(PushdataBytelength(v))), op::EQUAL] => {
+            v.as_slice()
+                .try_into()
+                .ok()
+                .map(|hash| ScriptKind::ScriptHash { hash })
```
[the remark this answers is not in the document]
str4d: The `try_into` is doing it; `hash` is a `[u8; 20]` in the enum variant.

src/script/mod.rs
```diff
@@ -266,6 +266,15 @@ impl PubKey {
...
     }
 
+    /// Returns true iff this script only contains `PushData`s.
+    ///
+    /// This is equivalent to checking that all opcodes are `<= OP_16`.
+    pub fn is_push_only(opcodes: &[Opcode]) -> bool {
+        opcodes
+            .iter()
+            .all(|opcode| matches!(opcode, Opcode::PushValue(_)))
+    }
```
sellout, line 276: This is an interesting case, only because script validation doesn’t ever care about `script::PubKey` being push-only.

The way to do this with `script::Sig` is something like
```rust
impl Sig<Opcode> {
  pub fn ensure_push_only(&self) -> Option<Sig<PushValue>> {…}
}
```
but `script::PubKey` isn’t parameterized in the same way. There are also some non-`pub` functions defined in a `slice` module for things that are common between `script::Sig` and `script::PubKey`.

I wonder if adding
```rust
impl PubKey {
  pub fn ensure_push_only(&self) -> Option<Sig<PushValue>> {…}
}
```
except that where you’re using this, you’re pattern matching, so you really do have a `&[Opcode]`.

In any case, I’d like something like `Meh<Opcode> -> Option<Meh<PushValue>>` to give us a usable value afterward.

src/standard.rs
```diff
@@ -0,0 +1,93 @@
...
+                .map(|op| match op {
+                    Opcode::PushValue(data) => data.value().filter(|data| {
+                        // Equivalent to `CPubKey::GetLen`
+                        let expected_len = match data.first() {
+                            Some(2 | 3) => crate::external::pubkey::PubKey::COMPRESSED_SIZE,
+                            Some(4 | 6 | 7) => crate::external::pubkey::PubKey::SIZE,
+                            _ => 0,
+                        };
+                        // Equivalent to `CPubKey::ValidSize`
+                        !data.is_empty() && expected_len == data.len()
```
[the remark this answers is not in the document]
str4d: I chose not to do this because the `zcashd` solver does not, and I was not yet sure whether the `zcashd` solver's output is necessary to preserve for compatibility. If we did do full parsing here then we'd be only matching a subset of possible scripts, which wouldn't be consensus-compatible but given that the `zcashd` solver was in `script/standard.h` it might not be used in a consensus-critical way.

src/standard.rs
```diff
@@ -0,0 +1,93 @@
...
+};
+
+/// Parses a `script::PubKey` and detects standard scripts.
+///
+/// If successful, returns the script kind, and any relevant properties parsed from the
+/// script. For example, for a P2SH script, the result will contain the script hash; for
+/// P2PKH it will contain the key hash, etc.
+///
+/// Returns `None` if the script is non-standard.
+pub fn solver(script_code: &script::PubKey) -> Option<ScriptKind> {
```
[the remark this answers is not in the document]
str4d: This is a more general issue: here we do not always have a `script_pubkey` (from an output), but may instead have a `redeem_script` (extracted from an input's `script_sig`). My recollection is that Bitcoin uses `script_code` in newer places where they are handling "any Script code", but I don't have a reference for this to hand.

str4d:
Rebased on #262 and addressed some comments.

str4d:
[Rebased on `master`](https://github.com/ZcashFoundation/zcash_script/compare/465f9058c7d23b2937d8f73d4a5b27d70524bf72..6a573a7f6e09bae3c4791ad2af53a329f5f1964c) after #262 merged.

daira (approved):
utACK

src/standard.rs
```diff
@@ -0,0 +1,93 @@
...
+        //
+        // So long as the script passes the `is_unspendable()` test, and all but the first
+        // byte passes the `is_push_only()` test, we don't care what exactly is in the
+        // script.
+        [op::RETURN, rest @ ..] if script::PubKey::is_push_only(rest) => Some(ScriptKind::NullData),
+
+        // Legacy Pay-to-Public-Key (P2PK), which only occurred in early Zcash coinbase outputs.
+        [Opcode::PushValue(PushValue::LargeValue(PushdataBytelength(v))), op::CHECKSIG] => {
+            (v.len() == PubKey::SIZE || v.len() == PubKey::COMPRESSED_SIZE)
+                .then(|| ScriptKind::PubKey { data: v.clone() })
```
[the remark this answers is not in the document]
str4d: `secp256k1` is no longer a hard dependendy.

src/solver.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        // Provably prunable, data-carrying output.
+        //
+        // So long as the script passes the `is_unspendable()` test, and all but the first
+        // byte passes the `is_push_only()` test, we don't care what exactly is in the
+        // script.
+        [op::RETURN, rest @ ..] => rest
+            .iter()
+            .map(|op| op.extract_push_value().ok().cloned())
```
sellout, line 44: Glad you’re using `extract_push_value` – it’s only ever collapsed to a `bool` in master, but this is the kind of thing I wanted to see.

src/solver.rs
```diff
@@ -0,0 +1,134 @@
...
+        _ => None,
+    }
+}
+
+/// Known kinds of standard scripts.
+pub enum ScriptKind {
+    /// A P2PKH script.
+    PubKeyHash {
+        /// The Hash160 of the public key.
+        hash: [u8; 20],
```
sellout, line 106: Should we have `type Hash160 = [u8; 20]` somewhere? It’s also used in the `pattern` module.

src/solver.rs
```diff
@@ -0,0 +1,134 @@
...
+                    data: v.clone().to_vec().try_into().expect("bounds checked"),
+                }
+            })
+        }
+
+        // Pay-to-Multi-Signature (P2MS)
+        [Opcode::PushValue(PushValue::SmallValue(required)), pubkeys @ .., Opcode::PushValue(PushValue::SmallValue(keys)), op::CHECKMULTISIG] =>
+        {
+            match (u8::try_from(required.to_num()), u8::try_from(keys.to_num())) {
+                (Ok(required @ 0..=16), Ok(keys @ 0..=16)) => {
```
sellout, line 64: In standard.cpp, `IsSmallInteger` checks that the value is at least 1.
```suggestion
                (Ok(required @ 1..=16), Ok(keys @ 1..=16)) => {
```

src/solver.rs
```diff
@@ -0,0 +1,134 @@
+//! Solvers for detecting known script kinds.
+
+use bounded_vec::BoundedVec;
+
+use crate::{
+    external::pubkey::PubKey,
```
str4d, line 6: Found a bug while updating https://github.com/zcash/librustzcash/pull/1721: this is now gated off, but we need it for its constants (which @sellout made associated constants somewhere in #174).

src/solver.rs
```diff
@@ -0,0 +1,134 @@
...
+                                let pubkey_bytes = data.value();
+                                // Equivalent to `CPubKey::GetLen`
+                                let expected_len = match pubkey_bytes.first() {
+                                    Some(2 | 3) => PubKey::COMPRESSED_SIZE,
+                                    Some(4 | 6 | 7) => PubKey::SIZE,
+                                    _ => 0,
+                                };
+                                // Equivalent to `CPubKey::ValidSize`
+                                (!pubkey_bytes.is_empty() && expected_len == pubkey_bytes.len())
+                                    .then(|| pubkey_bytes.try_into().expect("bounds checked"))
```
sellout, line 79: It’d be nice to avoid the lurking bug here – the `!pubkey_bytes.is_empty()` is only to protect against the `0` that represents an invalid result in the `match`.
```suggestion
                                // Equivalent to `CPubKey::GetLen`
                                let expected_len = match pubkey_bytes.first() {
                                    Some(2 | 3) => Some(PubKey::COMPRESSED_SIZE),
                                    Some(4 | 6 | 7) => Some(PubKey::SIZE),
                                    _ => None,
                                };
                                // Equivalent to `CPubKey::ValidSize`
                                expected_len
                                    .filter(|expected| expected == pubkey_bytes.len())
                                    .map(|_| pubkey_bytes.try_into().expect("bounds checked"))
```
or
```rust
                                (expected_len == Some(pubkey_bytes.len()))
                                    .then(|| pubkey_bytes.try_into().expect("bounds checked"))
```

src/solver.rs
str4d: Opened #266 for updating CI to catch this kind of issue going forward.

src/solver.rs
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/// Parses a `script::PubKey` and detects standard scripts.
+///
+/// If successful, returns the script kind, and any relevant properties parsed from the
+/// script. For example, for a P2SH script, the result will contain the script hash; for
+/// P2PKH it will contain the key hash, etc.
+///
+/// Returns `None` if the script is non-standard.
+pub fn standard(script_code: &script::PubKey) -> Option<ScriptKind> {
```
sellout, line 19: To @nuttycom’s comment before, this could be `script_code: &script::Component<Opcode>`. More verbose, but perhaps more correct, since this could also be a redeem script (`script::Redeem` – which all have the same underlying type).

I don’t have a preference, though.

sellout (changes requested):
I think the only blocking thing is the `IsSmallInteger` change.

src/solver.rs
str4d: Could be a useful refactor, but for a subsequent PR, and it should be `Hash160Output` or similar (as it's not the hash function itself).

src/solver.rs
str4d: Indeed, and I kept the `!pubkey_bytes.is_empty()` intentionally to match the C++ code for easier review. I'll add a follow-on commit that does the oxidation.

src/solver.rs
str4d: Good catch, thanks!

PR #254, opened 2025-08-18
Partition error types

sellout:


src/interpreter.rs
```diff
@@ -798,155 +789,88 @@ fn eval_operation(
...
+        })?),
 
         OP_2DROP => {
-            if stack.len() < 2 {
-                return Err(Error::InvalidStackOperation(None));
-            }
-
-            stack.pop()?;
-            stack.pop()?;
+            stack.rremove(1).and_then(|_| stack.pop())?;
```
nuttycom, line 821: The construction here is correct but nonobvious; it might be worth a comment.

src/interpreter.rs
```diff
@@ -64,18 +64,6 @@ pub enum Error {
     #[error("verify operation failed")]
     Verify,
 
-    #[error("equal verify operation failed")]
```
nuttycom: Is the distinction between these really not relevant? I could see it providing some useful information to someone in the process of iteratively developing a script.

src/interpreter.rs
```diff
@@ -1233,79 +1233,93 @@ impl SignatureChecker for CallbackTransactionSignatureChecker<'_> {
...
+/// Additional validation for spend-to-script-hash transactions:
+fn eval_p2sh<F>(
+    data_stack: Stack<Vec<u8>>,
+    payload: &mut F::Payload,
+    stepper: &F,
+) -> Result<Stack<Vec<u8>>, script::Error>
+where
+    F: StepFn,
+{
+    // stack cannot be empty here, because if it was the P2SH HASH <> EQUAL scriptPubKey would
```
nuttycom, line 1261: Why is it that we no longer need the push-only check here? The commit message on this commit is not very informative with respect to this change. Is it that this is only invoked from locations where the check has already been performed?

nuttycom (dismissed):
utACK 6dfcc81734131d423449de8312d6d88f0bac7147 with nonblocking questions.

src/interpreter.rs
sellout: The commit before this pairs each `interpreter::Error` with the opcode that triggered it, so the distinction remains.

At the level of interpreting a single opcode, you know which opcode triggered the error, so you don’t need the distinction there.

src/interpreter.rs
sellout: The check is still done, but it remains in `verify_script`. This is because it applies to the `Sig` component, not the `Redeem` component.

src/interpreter.rs
```diff
@@ -265,6 +265,15 @@ impl<T> Stack<T> {
         Stack(vec![])
     }
 
+    fn check_len(&self, min: usize) -> Result<(), Error> {
```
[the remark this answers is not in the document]
sellout: `NonZeroUsize` seems reasonable. I’m also unhappy with the subtraction here, but didn’t find a better way to approach it.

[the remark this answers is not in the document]
sellout: Oh wait, it’s safe to call `check_length` where `min` is 0, because the subtraction only happens if `min <= len` is false, and 0 is `<=` every usize. I.e., this checks that the length is at least `min`, and it’s always at least 0.

nuttycom (approved):
re-utACK e3a63f7da969af699c0eddeca098bd3df3b493d9

sellout:
Ok, it actually fails while building `zcash_script` _for_ fuzzing, not while fuzzing. So I’m guessing some of the changes here had an impact on the instrumentation.

sellout:
I tried adding `--jobs 1`[^1], but that just made it take longer before failing (I think the OOM happens solely when compiling zcash_script/src/lib.rs, so the lack of parallelism just delayed us getting there).

[^1]: Indirectly – `cargo fuzz build` doesn’t have a `--jobs` option, so I create a .cargo/config.yaml containing `build.jobs` in the CI job.

src/interpreter.rs
```diff
@@ -798,155 +789,88 @@ fn eval_operation(
...
 
         OP_PICK | OP_ROLL => {
             // (xn ... x2 x1 x0 n - xn ... x2 x1 x0 xn)
             // (xn ... x2 x1 x0 n - ... x2 x1 x0 xn)
-            if stack.len() < 2 {
-                return Err(Error::InvalidStackOperation(None));
-            }
-            let n = u16::try_from(num::parse(stack.rget(0)?, require_minimal, None)?)
+            stack.check_len(2)?;
+            let n = usize::try_from(num::parse(stack.rget(0)?, require_minimal, None)?)
```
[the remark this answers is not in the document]
sellout: Yeah, the stack and altstack can only ever have 1k elements between them and that’s checked after each operation.

PR #253, opened 2025-08-18
Partition `script` module

sellout:
Blocked by #224.

PR #255, opened 2025-08-20
change: Make script module public

upbqdn:
## Motivation

- https://github.com/zcash/librustzcash/pull/1893 uses [zcash_script::script::serialize_num](https://github.com/ZcashFoundation/zcash_script/blob/master/src/script.rs?rgh-link-date=2025-08-20T19%3A18%3A06Z#L407) here: https://github.com/zcash/librustzcash/blob/5ebd90259d90184dc8159cd602e572ec780d7ea4/zcash_transparent/src/bundle.rs#L255.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,7 @@ mod external;
 pub mod interpreter;
 pub mod op;
 pub mod pv;
-mod script;
+pub mod script;
 pub mod script_error;
 pub mod signature;
 mod zcash_script;


sellout:
> @sellout do you think `script` is stable enough to make it public?

I don’t think so. There are a few more open PRs to get in (#253, #254, #209), and then more PRs to open to catch up to the branch that @str4d and you have been working on.

#254 does make `script` public, but it looks pretty different at that point, and the coming PRs make more changes.

If you’re using the `ZcashScript` trait, then things are pretty stable. However, if that were the case, I don’t think `script` would need to be public. These coming PRs have a nicer Rust API, whereas `ZcashScript` is to maintain complete compatibility with the C++.

@upbqdn I’m happy to talk about what the changes look like. I don’t think any of it will be hard to adopt, whatever your use case, but it may help you reduce some churn if you know what’s coming, and what it’ll look like.

upbqdn:
Thanks for the explanation, sellout. Closing this in favor of the other PRs.

PR #248, opened 2025-07-25
Expose `script_sig` opcodes

str4d:
Only the last commit is relevant; it is based on the rest of @sellout's in-progress changes.

PR #256, opened 2025-08-21
add support for 'asm' display of scripts

conradoplg:
This is based on the same commit that https://github.com/ZcashFoundation/zcash_script/pull/248 is based on; only the last commit is relevant.

We need the "asm" encoding of the scripts to [finish the implementation of getrawtransaction](https://github.com/ZcashFoundation/zebra/issues/9330).

This PR does that by using the `Display` trait. Not sure if that was the best approach, let me know if you think it should be a separate trait.


sellout:
Master is stabilizing for 0.4.0. Now is a good time to resolve conflicts.

conradoplg:
> Master is stabilizing for 0.4.0. Now is a good time to resolve conflicts.

Thank you for the warning, I'll have it ready by tomorrow, I just noticed one thing I need to fix.

zcash_script/src/script/mod.rs
```diff
@@ -417,6 +439,30 @@ impl Evaluable for Code {
...
+        for r in self.parse_strict(&interpreter::Flags::StrictEnc) {
+            match r {
+                Ok(op) => v.push(op.to_asm(attempt_sighash_decode && !is_unspendable)),
+                Err(_) => {
+                    v.push("[error]".to_string());
+                    break;
+                }
+            }
+        }
+        v.join(" ")
```
sellout, line 455: This didn’t come out quite as nicely as I had hoped, because it short-circuits while preserving the result thus far, but just an option.

```suggestion
        let is_unspendable = self.is_unspendable();
        self.parse(&interpreter::Flags::StrictEnc)
            .try_fold(
                String::new(),
                |mut s, r| r
                    .map(|op| s.push_str(op.to_asm(attempt_sighash_decode && !is_unspendable) + " "))
                    .map_err(|_| s.push_str("[error]")),
            )
            .unwrap_or_else(|x| x)
            .trim_end()
```

zcash_script/src/opcode/mod.rs
```diff
@@ -1433,3 +1546,19 @@ mod tests {
...
+            Opcode::Control(c) => c.to_asm(attempt_sighash_decode),
+        }
+    }
+}
+
+impl core::fmt::Display for Opcode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.to_asm(false))
+    }
+}
```
sellout, line 1564: I think these should be in lib.rs

zcash_script/src/lib.rs
sellout, line 1: Should probably have an
```rust
impl<T: Asm, U: Asm> Asm for Script<T, U>
```
which looks like the one for `script::Raw`.

And that implies you’ll also need an
```rust
impl<T: Asm> Asm for Component<T>
```
in script.rs, which could share most of the `script::Code` implementation (there’s a `script::iter` mod, where I’ve put some of the common function bodies … I don’t love it, so open to suggestions there).

zcash_script/src/script/mod.rs
```diff
@@ -417,6 +439,30 @@ impl Evaluable for Code {
...
+    }
+}
+
+/// Implement Display for any type that implements Asm
```
sellout, line 459: This comment looks like it was for a generic impl that doesn’t exist any more.

zcash_script/src/script/mod.rs
```diff
@@ -364,6 +375,17 @@ impl Code {
...
+
+    /// Returns whether the script is guaranteed to fail at execution,
+    /// regardless of the initial stack.
+    fn is_unspendable(&self) -> bool {
+        const MAX_SCRIPT_SIZE: usize = 10000;
+        self.parse().next()
+            == Some(Ok(opcode::PossiblyBad::Good(Opcode::Operation(
+                opcode::Operation::OP_RETURN,
+            ))))
+            || self.0.len() > MAX_SCRIPT_SIZE
```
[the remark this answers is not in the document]
conradoplg: Thanks, good suggestion, I missed `Self::MAX_SIZE`.

Yep, the only goal here is to implement https://github.com/zcash/zcash/blob/2352fbc1ed650ac4369006bea11f7f20ee046b84/src/script/script.h#L617-L620. I updated the docs.

zcash_script/src/script/mod.rs
```diff
@@ -417,6 +439,30 @@ impl Evaluable for Code {
     }
 }
 
+impl Asm for Code {
+    fn to_asm(&self, attempt_sighash_decode: bool) -> String {
+        let mut v = Vec::new();
+        let is_unspendable = self.is_unspendable();
+        for r in self.parse_strict(&interpreter::Flags::StrictEnc) {
```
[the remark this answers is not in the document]
conradoplg: You're right - I used `parse_strict()` because of `StrictEnc` but that does not make sense here, I just need it for checking signature format which is done elsewhere. I changed it to use `parse()`

zcash_script/src/script/mod.rs
conradoplg: That wasn't working and I couldn't make a similar idea work. I think it's confusing mixing functional and imperative code, but even if I changed it to concatenate strings in `try_fold()`, it doesn't solve the tricky part is that we need to stop it in the _next_  iteration after an error. The only idea I can come up is to return a flag along with the value in `try_fold()` to indicate that we should stop at the next iteration, but at that point the code would look much more confusing than the current imperative version.

zcash_script/src/script/mod.rs
conradoplg: Oops. Thanks for catching it. Deleted the comment

zcash_script/src/opcode/mod.rs
conradoplg: Moved.

conradoplg (commented):
Addressed comments in bdcf320774

zcash_script/src/opcode/mod.rs
```diff
@@ -1028,6 +1063,84 @@ impl From<&PushValue> for Vec<u8> {
     }
 }
 
+impl Asm for Operation {
+    fn to_asm(&self, _attempt_sighash_decode: bool) -> String {
```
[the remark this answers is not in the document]
sellout: I’m not opposed, but doesn’t have to happen now.

conradoplg:
> Also, any feedback you have on how it was working with the crate / API would be great.

Ah, forgot to reply this - the API is good, just a bit confusing, it's not immediately clear the difference between Script / Code etc. But I think we just need some high level documentation to address that.

str4d:
This PR conflicted with #263, and both `zcash_script 0.5.0` (and now `zcash_script 0.4.1`) break the no-std support that was released in `zcash_script 0.4.0`.

PR #262, opened 2025-09-17
Stronger script types

sellout:
`script::Raw` is the type for scripts read from the chain, while `Script` is a new type for authoring scripts (e.g., via PCZTs).

`Script` takes two type parameters – one for the opcode type for each of its components. `Script` defaults to `Script<PushValue, Opcode>`, which is the type that is generally desired for authoring. There is also `script::Component<T>`, with `type script::Sig = Component<PushValue>` and `type script::PubKey = Component<Opcode>`.

Blocked by #261.


sellout:
This doesn’t contain everything from the branch @str4d has been using for PCZTs, but I think it contains everything necessary for current usage. Getting the remainder in shape on top of everything else was a bit of a slog, so I figured if I get this up ASAP, we can get PCZTs migrated over, and also catch anything that _is_ missing.

@str4d – as mentioned in the description, the `Script` types are a bit different than they were at the point you’re working on. I can help with the migration if you want.

src/opcode/mod.rs
```diff
@@ -38,6 +38,31 @@ pub enum Error {
...
+/// Definitions needed for evaluation of script types.
+pub trait Evaluable {
+    /// The length in bytes of this script value. This can be more efficient than
+    /// `self.to_bytes().len()`.
+    fn byte_len(&self) -> usize;
+
+    /// Convert a script value into the bytes that would be included in a transaction.
+    fn to_bytes(&self) -> Vec<u8>;
+
+    fn restrict(pb: PossiblyBad) -> Result<Self, script::Error>
```
str4d, line 50: This either needs documentation, or (more likely) should be:
```suggestion
    #[doc(hidden)]
    fn restrict(pb: PossiblyBad) -> Result<Self, script::Error>
```
(because `PossiblyBad` is not nameable in the public API).

src/opcode/mod.rs
```diff
@@ -38,6 +38,31 @@ pub enum Error {
...
+    /// Evaluate the provided script value.
+    fn eval(
+        &self,
+        flags: interpreter::Flags,
+        script: &script::Code,
+        checker: &dyn interpreter::SignatureChecker,
+        state: interpreter::State,
+    ) -> Result<interpreter::State, interpreter::Error>;
+
+    fn extract_push_value(&self) -> Result<&PushValue, script::Error>;
```
str4d, line 63: Needs documentation.

src/lib.rs
```diff
@@ -508,26 +567,28 @@ pub mod testing {
...
             pv::push_value(&<[u8; 0x47]>::from_hex("3044022013e15d865010c257eef133064ef69a780b4bc7ebe6eda367504e806614f940c3022062fdbc8c2d049f91db2042d6c9771de6f1ef0b3b1fea76c1ab5542e44ed29ed801").expect("valid sig")).expect("fits into a PushValue"),
             push_script(&REDEEM_SCRIPT).expect("fits into a PushValue"),
-        ].map(Opcode::PushValue));
+        ]);
         /// The combined script used for the static test case.
-        pub static ref SCRIPT: script::Raw<'static> =
-            script::Raw::from_raw_parts(&SCRIPT_SIG, &SCRIPT_PUBKEY);
+        pub static ref SCRIPT: script::Raw =
+            script::Raw::from_raw_parts(SCRIPT_SIG.to_bytes(), SCRIPT_PUBKEY.to_bytes());
+        pub static ref REAL_SCRIPT: Script =
```
str4d, line 590: Needs documentation.

src/opcode/mod.rs
```diff
@@ -106,7 +131,7 @@ impl PushValue {
         }
     }
 
-    pub fn eval(
+    pub fn eval_(
```
str4d, line 134: Needs documentation. It will be confusing to have both `.eval()` and `.eval_()` in the public API on the same type, so also check if this can be private, and if so:
```suggestion
    fn eval_(
```
(or `pub(crate)` if necessary).

src/opcode/mod.rs
```diff
@@ -950,12 +1010,29 @@ impl PossiblyBad {
             PossiblyBad::Bad(_) => Err(vec![interpreter::Error::BadOpcode]),
         }
     }
+}
+
+impl Evaluable for PossiblyBad {
```
str4d, line 1015: A problem I ran into while trying to fix the issues with https://github.com/zcash/librustzcash/pull/1721 is that `PossiblyBad` is a `pub struct`, but is not nameable in the public API. That means even with this impl, it was impossible for me to construct a `script::Component<PossiblyBad>` to use as an alternative for `script::PubKey`, and due to not knowing which of the two possibilities was intended here (that `PossiblyBad` be in the public API, or an internal-only detail), I couldn't rely on it and had to switch back to plain `Vec<u8>` for transaction parsing.

As part of figuring out the public API, please ensure that structs like `PossiblyBad` are only `pub` if they are either part of the nameable public API, or are clearly documented as to why they are `pub` (e.g. trait shenanigans).

src/opcode/push_value.rs
```diff
@@ -28,6 +28,15 @@ impl LargeValue {
 
     pub const MAX_SIZE: usize = 520; // bytes
 
+    pub fn byte_len(&self) -> usize {
```
str4d, line 31: Needs documentation.

src/script/iter.rs
```diff
@@ -0,0 +1,144 @@
...
+                            None => u32::from(interpreter::MAX_PUBKEY_COUNT),
+                        }
+                    }
+                    _ => 0,
+                },
+                _ => 0,
+            },
+            (_, _) => 0,
+        })
+        .sum()
```
str4d, line 143: Possible bug: In the previous code, this returned `Err(n)`, short-circuiting the iterator (and its count) at the first bad opcode. In this code, we treat bad opcodes as 0, but keep counting.

If the previous behaviour was consensus-critical (which I suspect it was), change this back to using a mechanism with the early-return behaviour.

src/script/mod.rs
```diff
@@ -115,6 +116,125 @@ impl From<opcode::Error> for Error {
...
+
+    /// Called by `IsStandardTx` and P2SH/BIP62 VerifyScript (which makes it consensus-critical).
+    fn is_push_only(&self) -> bool;
+}
+
+/// A script component is used as either the script sig or script pubkey in a script. Depending on
+/// `T` (which generally needs an `opcode::Evaluable` instance), it has different properties:
+///
+/// - `PossiblyBad` – used for scripts that need to go through consensus-compatible verification
+///   (that is, read from the chain)
```
str4d, line 146: Now that I see this documentation, it suggests that `PossiblyBad` being unnameable in the public API is the bug.

src/script/mod.rs
```diff
@@ -115,6 +116,125 @@ impl From<opcode::Error> for Error {
...
+/// A script pubkey has any `Opcode`.
+pub type PubKey = Component<Opcode>;
+
+impl<T: opcode::Evaluable> Component<T> {
+    /// This parses an entire script. This is stricter than the incremental parsing that is done
+    /// during `verify_script`, because it fails on unknown opcodes no matter where they occur.
+    /// I.e., this is useful for validating and analyzing scripts before they are put into a
+    /// transaction, but not for scripts that are read from the chain, because it may fail on valid
+    /// scripts.
+    pub fn parse(raw_script: &Code) -> Result<Self, Error> {
```
str4d, line 165: If `PossiblyBad` is added to the public API, then I think this documentation would be incorrect for it (because it wouldn't fail on unknown opcodes, just on incorrect structural encoding).

Ideally we'd have two parsing methods: `Component<PossiblyBad>::parse_lax` which only performs structural checks, and `Component<not PossiblyBad>::parse_strict` which does what this parser does. Given the lack of time to engineer that solution (which would likely add complexity due to needing an additional trait layer to express the `not PossiblyBad`), that is explicitly deferred to a future issue.

For this PR just alter the documentation to clarify that it is only stricter when `T` is not `PossiblyBad`.

src/script/mod.rs
```diff
@@ -178,100 +298,65 @@ impl Code<'_> {
     }
 
     /// Produce an [`Opcode`] iterator from [`Code`].
+    ///
+    /// It returns `Err` with the length if the script is too long.
     pub fn parse(&self) -> Parser<'_> {
```
str4d, line 303: What is `It` here? The method doesn't return a `Result`; are you saying that the returned `Parser` will contain a single `Err` item in this case?

src/script/mod.rs
```diff
@@ -178,100 +298,65 @@ impl Code<'_> {
...
+
+impl Evaluable for Code {
+    /// Get the byte length of this script sig.
+    fn byte_len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Convert a sequence of `Opcode`s to the bytes that would be included in a transaction.
+    fn to_bytes(&self) -> Vec<u8> {
+        self.0.to_vec()
```
str4d, line 330: After the change to make `Code` own its bytes, this would be clearer:
```suggestion
        self.0.clone()
```

str4d (changes requested):
Reviewed 3b78e4f29333271ee35a2acbf78bc89a85a44608.

src/script/mod.rs
```diff
@@ -115,6 +110,125 @@ impl From<opcode::Error> for Error {
...
+///   on the chain
+/// - `PushValue` – used for push-only script sigs that are authored to be placed on the chain
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Component<T>(pub Vec<T>);
+
+/// A script sig has only `PushValue` elements.
+pub type Sig = Component<PushValue>;
+
+/// A script pubkey has any `Opcode`.
+pub type PubKey = Component<Opcode>;
```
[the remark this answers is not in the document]
str4d: Having now used `Component` in https://github.com/zcash/librustzcash/pull/1721, I agree that it's fine for `Component` to be a pure newtype wrapper. AFAICT we are not relying on `Component` to enforce any correctness requirements on its contents (it is either enforced by `T`, or only checked at runtime during script evaluation). So we don't need #248.

src/signature.rs
```diff
@@ -190,6 +193,12 @@ pub struct Decoded {
     hash_type: HashType,
 }
 
+/// There’s not such thing as a decoded signature unless the `eval` feature is enabled.
```
str4d, line 196:
```suggestion
/// There’s not such thing as a decoded signature unless the `signature-validation` feature is enabled.
```

str4d (changes requested):
Reviewed db1d599ff4480944543393faa954402e0965eddf. We still need to figure out whether the sigop counting difference is consensus-critical.

src/lib.rs
```diff
@@ -149,6 +180,33 @@ impl From<&Opcode> for Vec<u8> {
     }
 }
 
+/// A Zcash script consists of a sig and a pubkey.
+///
+/// - Script<opcode::PossiblyBad, opcode::PossiblyBad> – from the chain
+/// - Script<opcode::PushValue, opcode::Structured> – authoring sig_push_only
+/// - Script<opcode::Structured, opcode::Structured> – authoring non-sig_push_only
+///
+/// Until `Structured` is added, we use `Opcode` in its place.
```
daira, line 189: It is not necessarily clear from this documentation that the first type parameter restricts the sig and the second restricts the pubkey, so that should be said explicitly.

src/opcode/mod.rs
daira: It's not obvious to me what this method does, so I think it should be documented whether or not it is usable in the public API. (Note that doc comments are allowed on an item with `#[doc(hidden)]`. Indeed there is an unstable `--document-hidden-items` flag to Rustdoc that shows them, although it is [currently not exposed by `cargo doc`](https://github.com/rust-lang/cargo/issues/7766#issuecomment-803731540).)

src/script/iter.rs
```diff
@@ -0,0 +1,118 @@
...
+/// they are of the form ... OP_N CHECKMULTISIG ...
+pub fn sig_op_count<T: Into<opcode::PossiblyBad> + opcode::Evaluable>(
+    mut iter: impl Iterator<Item = Result<T, opcode::Error>>,
+    accurate: bool,
+) -> u32 {
+    iter.try_fold((0, None), |(sum, last_opcode), opcode| {
+        opcode.map_err(|_| sum).map(|op| {
+            (
+                sum + op.sig_op_count(last_opcode),
+                if accurate { Some(op.into()) } else { None },
```
daira, line 112:
```suggestion
                accurate.then(op.into()),
```

src/opcode/mod.rs
```diff
@@ -106,7 +143,11 @@ impl PushValue {
         }
     }
 
-    pub fn eval(
+    /// The internal implementation of `Evaluable::eval`. `eval` has larger parameters, because it
+    /// needs to be general. This only has the parameters required by the implementation. The
+    /// `Evalable` `impl` extracts the necessary components from the larger parameters and passes
```
daira, line 148:
```suggestion
    /// `Evaluable` `impl` extracts the necessary components from the larger parameters and passes
```

src/opcode/mod.rs
```diff
@@ -106,7 +143,11 @@ impl PushValue {
...
+    /// them through.
+    fn eval_(
```
daira, line 150: Consider renaming to `eval_internal`.

daira (approved):
utACK; remaining comments are non-blocking.

PR #264, opened 2025-09-19
Remove `enum_primitive` dependency

str4d:
`enum_primitive` was last updated 8 years ago, and it depends on `num-traits 0.1`. This causes a bunch of type confusion problems in projects that depend on `num-traits 0.2` (as that crate used the "compatibility trick" of re-exporting the `num-traits 0.2` API inside the last release of `num-traits 0.1`, which today causes ambiguous trait resolution errors).


str4d:
[Rebased](https://github.com/ZcashFoundation/zcash_script/compare/8dc114706ab424409dd84428d7671c9e955c3c86..03bf5b5141482c200e76cda70ac07e06900736f0) on #247 again after it was rebased.

src/opcode/mod.rs
```diff
@@ -1177,3 +1177,144 @@ impl From<Disabled> for u8 {
...
+            Disabled::OP_OR,
+            Disabled::OP_XOR,
+            Disabled::OP_2MUL,
+            Disabled::OP_2DIV,
+            Disabled::OP_MUL,
+            Disabled::OP_DIV,
+            Disabled::OP_MOD,
+            Disabled::OP_LSHIFT,
+            Disabled::OP_RSHIFT,
+            Disabled::OP_CODESEPARATOR,
```
daira, line 1411: I have not checked that this is exactly the disabled opcodes.

daira (approved):
utACK

PR #267, opened 2025-09-22
Fix the fuzzer

sellout:
`script::Raw` had its type change, but the fuzzer didn’t get updated at the time.


str4d (approved):
utACK 072162d967b64de8d705a71a8955dfa8477c490c

PR #268, opened 2025-09-22
Expand `signature-validation` feature

sellout:
`CallbackTransactionSignatureChecker` is now only defined if `signature-validation` is enabled. Otherwise, callers are forced to use `NullSignatureChecker`, making it clearer that validation won’t happen.


src/lib.rs
```diff
@@ -481,25 +481,50 @@ pub struct ComparisonInterpreter<T, U> {
     second: U,
 }
 
+/// Creates an `interpreter::SignatureChecker`. The actual type depends on whether the
+/// `signature-validation` feature is enabled. If it’s not, it will create a checker that always
+/// fails.
+#[cfg(feature = "signature-validation")]
+pub fn create_signature_checker(
```
daira, line 488: This is kind-of missing the point of my [original comment](https://github.com/ZcashFoundation/zcash_script/pull/262#discussion_r2367837703), which was that the semantics of an API (rather than its presence) shouldn't normally be dependent on a feature flag.

src/lib.rs
```diff
@@ -481,25 +481,50 @@ pub struct ComparisonInterpreter<T, U> {
...
+    }
+}
+#[cfg(not(feature = "signature-validation"))]
+pub fn create_signature_checker(
+    _sighash: interpreter::SighashCalculator,
+    _lock_time: i64,
+    _is_final: bool,
+) -> impl interpreter::SignatureChecker {
+    interpreter::NullSignatureChecker()
+}
```
daira, line 506:
```suggestion
```

src/lib.rs
```diff
@@ -481,25 +481,50 @@ pub struct ComparisonInterpreter<T, U> {
...
 
+/// Creates an `interpreter::SignatureChecker`. The actual type depends on whether the
+/// `signature-validation` feature is enabled. If it’s not, it will create a checker that always
+/// fails.
+#[cfg(feature = "signature-validation")]
+pub fn create_signature_checker(
+    sighash: interpreter::SighashCalculator,
+    lock_time: i64,
+    is_final: bool,
+) -> impl interpreter::SignatureChecker + '_ {
```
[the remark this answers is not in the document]
daira: If this were not public (as in my suggestion), then this would be fine. I think you're right that it's problematic as is.

src/lib.rs
sellout: Yeah, this was me trying to keep the tests intact, when really I should have gone the other way. `create_signature_checker` is now gone. `cxx_rust_comparison_interpreter` is conditionalized on the `signature-validation` feature (because they’re really not comparable if signature validation is disabled), and I’ve split a bunch of tests up that previously used the comparison interpreter to now just test the individual interpreters (the comparison interpreter is exercised by the arbitrary scripts tests).

sellout: I think this change addresses most of the comments.

libzcash_script/src/lib.rs
```diff
@@ -385,86 +387,126 @@ mod tests {
...
+            lock_time: lock_time.into(),
+            is_final,
+        })
+        .verify_callback(&SCRIPT, flags);
+
+        assert_eq!(ret, Ok(false));
+    }
+
+    #[test]
+    fn it_fails_on_invalid_sighash_with_null_checker() {
```
str4d, line 461: AFAICT this is identical to `it_fails_with_null_checker` and is not checking anything new. Remove this.

libzcash_script/src/lib.rs
```diff
@@ -385,86 +387,126 @@ mod tests {
...
+            lock_time: lock_time.into(),
+            is_final,
+        })
+        .verify_callback(&SCRIPT, flags);
+
+        assert_eq!(ret, Ok(false));
+    }
+
+    #[test]
+    fn it_fails_on_missing_sighash_with_null_checker() {
```
str4d, line 503: AFAICT this is identical to `it_fails_with_null_checker` and is not checking anything new. Remove this.

str4d (approved):
utACK abc410cd2bcebcb71b69eb46edbefde3e0d3e0c2 with notes.

daira (commented):
Post-hoc ACK

PR #274, opened 2025-09-24
Minimize the public API

sellout:


str4d:
CI failures are due to partal service disruption at Docker (https://www.dockerstatus.com/), nothing we can do about it, but they also don't matter here because dependencies aren't being altered.

PR #275, opened 2025-09-25
Release `zcash_script 0.4.0`, `libzcash_script 0.1.0`

str4d:
Given that our eventual plan is to end support for `libzcash_script`, my proposal is that we keep tagging the `zcash_script` release commits with the regular tags (so this would be `v0.4.0`), while the `libzcash_script` release tags are prefixed (so this would be `libzcash_script-v0.1.0`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,7 +736,7 @@ checksum = "589f6da84c646204747d1270a2a5661ea66ed1cced2631d546fdfb155959f9ec"
 
 [[package]]
 name = "zcash_script"
-version = "0.3.2"
+version = "0.4.0"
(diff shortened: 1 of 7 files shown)


str4d:
CI failures are due to partal service disruption at Docker (https://www.dockerstatus.com/), nothing we can do about it.

str4d:
Published 670b41281963196f8c7add4fa957f8acb19a4e5b as `v0.4.0`

str4d:
Published 543c3790c3da6b3281025ff0c90e6f463d348d17 as `libzcash_script-v0.1.0`.

PR #304, opened 2026-05-29
make sig_op_count() match zcashd

conradoplg:
Fixes sig op count in the Rust path to match zcashd behaviour


arya2 (approved):
Looks good!

